PR #11, opened 2017-05-24
Make removeListener not crash (argument list typo)

vasklund:
`UrbanAirship.removeListener` has a typo in the argument list (`handler` instead of `listener`), which causes the method to throw when used.

This PR fixes that issue, as well as improving the documentation for the method (no `EmitterSubscription` is returned by removeListener as far as I can see).


crow:
Thanks for your contribution!

PR #14, opened 2017-06-03
Fixed addCustomEvent

timeout2x:
Adding customEvent with Dictionary (Object from javascript) triggers error since `runAction` expects string. 

```
[RCTConvert.m:58] JSON value '{
    ...
}' of type NSMutableDictionary cannot be converted to NSString
```

The next function in chain expects dictionary arguments and triggers another error:
```
[UAAddCustomEventAction acceptsArguments:] [Line 46] UAAddCustomEventAction requires a dictionary of event data.
```

Without this fix it is impossible to use `addCustomEvent` in ios.


ralepinski:
@timeout2x Thanks for the PR! I think there is a bit more work we need to do to fix this. Actions should be able to accept any value, not just string or dictionary. Maybe we can use "id". We will try to get a release today.

crow:
Released version 1.0.3 with the fix, thanks again @timeout2x

PR #16, opened 2017-06-09
[MB-2258] iOS Badging Support

gatlinhebert:
Adds the ability to set and get iOS badge counts.

### DIFF
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -402,6 +402,24 @@ class UrbanAirship {
   static isQuietTimeEnabled(): Promise<boolean> {
     return UrbanAirshipModule.isUserNotificationsEnabled();
   }
+
+  /**
+   * Sets the badge number for iOS.
+   *
+   * @param {number} badgeNumber specified badge to set.
+   */
+  static setBadgeNumber(badgeNumber: number) {
+    UrbanAirshipModule.setBadgeNumber(badgeNumber);
+  }
+
+  /**
+   * Gets the current badge number for iOS.
+   *
+   * @return {Promise.<number>} A promise with the result.
+   */
+  static getBadgeNumber(): Promise<number> {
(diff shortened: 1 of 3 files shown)


js/UrbanAirship.js
```diff
@@ -402,6 +402,24 @@ class UrbanAirship {
...
     return UrbanAirshipModule.isUserNotificationsEnabled();
   }
+
+  /**
+   * Sets the badge number for iOS.
+   *
+   * @param {number} badgeNumber specified badge to set.
+   */
+  static setBadgeNumber(badgeNumber: number) {
+    UrbanAirshipModule.setBadgeNumber(badgeNumber);
```
rlepinski: This will fail on Android. Should we call wrap it in a platform check?

linkmh:
:shipit:

js/UrbanAirship.js
```diff
@@ -402,6 +402,24 @@ class UrbanAirship {
...
+  }
+
+  /**
+   * Gets the current badge number for iOS.
+   *
+   * @return {Promise.<number>} A promise with the result.
+   */
+  static getBadgeNumber(): Promise<number> {
+    return UrbanAirshipModule.getBadgeNumber();
```
rlepinski: Should we implement this method on Android and return 0 everytime?

js/UrbanAirship.js
gatlinhebert: Would the following make sense? 
```
if (Platform.OS == 'ios') {
      UrbanAirshipModule.setBadgeNumber(badgeNumber);
} else {
      console.log('This feature is not supported on this platform.')
 }
```
Or is this something that should be handled within app code? 

```
if (Platform.OS == 'ios') {
  UrbanAirship.setBadgeNumber(1);
}
```

rlepinski:
👍

PR #25, opened 2017-07-21
[MB-2246] Adds the ability to dismiss the default message center.

gatlinhebert:
Adds the ability to dismiss showing message center and message center messages. Dismissing message center is useful for apps that use the displayMessageCenter, while individual message dismissal is used for when an app displays their own message displaying.

On Android, its hard to access the activity stack in a clean non terrible way. Instead we will send an intent to the activities to close themselves. If the activity is currently running it will  receive the new intent through `onNewIntent`, if its not running it will just close itself immediately before its displayed to the user.

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/CustomMessageCenterActivity.java
+++ b/android/src/main/java/com/urbanairship/reactnative/CustomMessageCenterActivity.java
@@ -11,7 +11,7 @@ public class CustomMessageCenterActivity extends MessageCenterActivity {
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        if (getIntent() != null && "CLOSE".equals(getIntent().getAction())) {
+        if (getIntent() != null && UrbanAirshipReactModule.CLOSE_MESSAGE_CENTER.equals(getIntent().getAction())) {
             finish();
             return;
         }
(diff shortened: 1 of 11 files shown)


android/src/main/java/com/urbanairship/reactnative/CustomMessageCenterActivity.java
```diff
@@ -0,0 +1,29 @@
...
+            finish();
+            return;
+        }
+    }
+
+    @Override
+    protected void onNewIntent(Intent intent) {
+        super.onNewIntent(intent);
+
+        if (intent != null && "CLOSE".equals(intent.getAction())) {
```
crow: Probably can pull "CLOSE" out as a constant like:
    private static final String CLOSE_MESSAGE_CENTER = "CLOSE";

crow:
🍔 manually tested on Android and iOS - looks good!

sample/AirshipSample/app.js
```diff
@@ -260,42 +270,50 @@ export default class AirshipSample extends Component {
...
-        <EnableLocationCell
-          locationEnabled={this.state.locationEnabled}
-          handleLocationEnabled={this.handleLocationEnabled}
-        />
-        <Text style={styles.instructions}>
-          Press Cmd+R to reload,{'\n'}
-          Cmd+D or shake for dev menu
-        </Text>
-      </View>
+      <View style={styles.backgroundContainer}>
```
crow: This is me adding the view hierarchy into a scroll view and adding the message button functionality

rlepinski:
👍

PR #29, opened 2017-08-16
Channel reg callbacks emit opt status event on opt status change

crow:
manually tested on Nexus 5x

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
@@ -1,0 +1,54 @@
+/* Copyright 2017 Urban Airship and Contributors */
+
+package com.urbanairship.reactnative;
+
+import android.content.Context;
+import android.content.SharedPreferences;
+
+import com.urbanairship.UAirship;
+import com.urbanairship.reactnative.events.NotificationOptInEvent;
+
+/**
+ * Stores shared preferences and checks preference-dependent state.
+ */
+public class ReactAirshipPreferences {
+
+    private static ReactAirshipPreferences sharedInstance = new ReactAirshipPreferences();
+
+    private SharedPreferences preferences;
+
+    private static final String SHARED_PREFERENCES_FILE = "com.urbanairship.reactnative";
+
+    protected static final String NOTIFICATIONS_OPT_IN_KEY = "NOTIFICATIONS_OPT_IN_KEY";
+
+    /**
+     * Returns the shared {@link ReactAirshipPreferences} instance.
+     *
+     * @return The shared {@link ReactAirshipPreferences} instance.
+     */
+    static ReactAirshipPreferences shared() {
+        return sharedInstance;
+    }
+
+    public SharedPreferences getPreferences (Context context) {
+        if (preferences == null) {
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Returns the shared {@link ReactAirshipPreferences} instance.
+     *
+     * @return The shared {@link ReactAirshipPreferences} instance.
+     */
+    static ReactAirshipPreferences shared() {
+        return sharedInstance;
+    }
+
+    public SharedPreferences getPreferences (Context context) {
```
rlepinski: make this private

android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
```diff
@@ -0,0 +1,54 @@
...
+        if (preferences == null) {
+            preferences = context.getSharedPreferences(SHARED_PREFERENCES_FILE, Context.MODE_PRIVATE);
+        }
+
+        return preferences;
+    }
+
+    public void checkOptInStatus (Context context) {
```
rlepinski: getOptInStatus*

crow: do we want to send an event as a side effect?

android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
crow: I can just remove it - I added it to expose the shared prefs file to the rest of the module, probably should have been protected

marc-scig:
Lets javadoc all the public things (as long as they aren't overrides)

android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -98,6 +99,17 @@ public void onHostPause() {
             public void onHostDestroy() {
 
             }
+
+            public void checkOptIn(Context context) {
+                boolean optIn = UAirship.shared().getPushManager().isOptIn();
```
marc-scig: is there any way to avoid repeating ourselves here? it looks like a 1:1 copy of the checkOptIn above

crow: Not unless we want the preferences class to do this which would be a side effect

android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
```diff
@@ -0,0 +1,57 @@
...
+
+        return preferences;
+    }
+
+    /**
+     * Saves opt in status in shared preferences.
+     * @param optIn The opt in state.
+     * @param context The application context.
+     */
+    public void setOptInStatus (boolean optIn, Context context) {
```
rlepinski: no whitespace between method name and params `setOptInStatus(`

marc-scig:
👍

rlepinski:
👍

PR #32, opened 2017-08-30
This is no longer needed in android

LandonS23:
This causes build Error:(25, 5) error: method does not override or implement a method from a supertype


rlepinski:
Going to go with PR #33 to keep support for older react versions. Thanks!

PR #52, opened 2017-10-26
Release 1.2.3

rlepinski:


marc-scig:
Added ryan's on-the-fly manifest change, the message center title was previously set on the wrong activity. All good now.

👍

Issue #54, opened 2017-11-06
Manage Delivered Notifications


AntiHate:
It would be great if we could have access to delivered notifications (Notifications Panel). To remove specific or all notifications. Or should we look for another library for that? Thanks!

Edit: I found this [WIX React native notifications](https://github.com/wix/react-native-notifications) but it'd nice to use just one lib for notifications.

![image](https://user-images.githubusercontent.com/2887867/32425885-5a0da5b4-c2d8-11e7-9ab1-b13a87e5b1ec.png)

rlepinski:
@AntiHate Remove all notifications is really easy, exposing individual notifications is a little harder. What APIs are you looking for?

rlepinski:
The API you provided is only offered for iOS 10+. The same API on Android can almost be achieved with API 23+, but accessing the original payload would be difficult for notifications posted outside of Urban Airship. 

https://developer.android.com/reference/android/app/NotificationManager.html#getActiveNotifications()

AntiHate:
Hi @rlepinski , Thankyou for your response. Right now I just need to cancel all the delivered notifications but it'd be great if we could get access to all delivered notifications data on both platforms so I'd able to remove specific notifications.

AntiHate:
Do you plan to offer Local Notifications API? or this is something out of this project scope?

AntiHate:
Thankyou @rlepinski for considering this. I'll look for it in the next release.

AntiHate:
@rlepinski You made it along with management support, that was quick 🥇 Thankyou so much!

rlepinski:
@AntiHate We will probably release early next week. I still need to verify everything works properly on iOS.

rlepinski:
Released in 1.3.0. Added:
- [clearNotification](https://docs.urbanairship.com/reference/libraries/react-native/latest/UrbanAirship.html#.clearNotification)
- [clearNotifications](https://docs.urbanairship.com/reference/libraries/react-native/latest/UrbanAirship.html#.clearNotifications)
- [getActiveNotifications](https://docs.urbanairship.com/reference/libraries/react-native/latest/UrbanAirship.html#.getActiveNotifications)

AntiHate:
@rlepinski  You are the best. Thankyou so much!

PR #56, opened 2017-11-10
Add notification management support

rlepinski:
Adds support to clear and get active notifications. Adds 3 methods:

- getActiveNotifications: Returns the current notifications in the notification center
- clearNotifications: Clears all notifications for the app.
- clearNotification(identifier): Clears a specific notification.

Android notification Ids are a unique combination of a tag and int id. To normalize the API between iOS and Android, Android will expose the notification Id as "id:tag" and then parse it to do the actual clear.

https://github.com/urbanairship/react-native-module/issues/54

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Version 1.3.0 - November 15, 2017
+=================================
+- Added APIs to manage active notifications.
+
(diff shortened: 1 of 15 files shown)


crow:
🍔 manually tested?

rlepinski:
Yep

PR #60, opened 2017-11-25
Fix Android event emission

### LINKED ISSUE #58: [Android] Emitted deep link events are sometimes not handled
**NOTE**: My app uses https://github.com/wix/react-native-navigation

Deep link events are always emitted, but sometimes not handled. Particularly when:
- The app is killed
- The app is in the background after a hardware back button exit

I managed to do a quick and dirty fix, by altering `EventEmitter.java`:
```java
final ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
if (reactContext != null && reactContext.hasActiveCatalystInstance()) {
    final Handler handler = new Handler();
    handler.postDelayed(new Runnable() {
        @Override
        public void run() {
            emit(reactContext, event.getName(), event.getBody());
        }
    }, 3500);
} else if (event.isCritical() || reactInstanceManager.hasStartedCreatingInitialContext()) {
    reactInstanceManager.addReactInstanceEventListener(new ReactInstanceManager.ReactInstanceEventListener() {
        public void onReactContextInitialized(final ReactContext reactContext) {
            final Handler handler = new Handler();
            handler.postDelayed(new Runnable() {
                @Override
                public void run() {
                    emit(reactContext, event.getName(), event.getBody());
                }
            }, 3500);
            reactInstanceManager.removeReactInstanceEventListener(this);
        }
    });
}
```
I delay the emission to ensure that the JS is ready to handle the event.

A better way to fix this issue is to keep track whether emitted events have been handled, and if not, reemitting them after a certain delay.

bevkoski:
Fixes #58.

I was aiming to make it as similar as possible to the current iOS implementation.

Because of https://github.com/facebook/react-native/blob/a93b7a2da0849f15708afef1ff99022d8cc55b14/Libraries/EventEmitter/NativeEventEmitter.js#L37 I had to add two additional `@ReactMethod`s:
- `increaseListenerCount()`
- `decreaseListenerCount()`

### DIFF
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -357,7 +357,7 @@ class UrbanAirship {
    */
   static addListener(eventName: UAEventName, listener: Function): EmitterSubscription {
     if (Platform.OS === 'android') {
-      UrbanAirshipModule.addListener();
+      UrbanAirshipModule.increaseListenerCount();
     }
     var name = convertEventEnum(eventName);
     return AirshipNotificationEmitter.addListener(name, listener);
@@ -372,7 +372,7 @@ class UrbanAirship {
    */
   static removeListener(eventName: AirshipEventName, listener: Function) {
     if (Platform.OS === 'android') {
-      UrbanAirshipModule.removeListener();
+      UrbanAirshipModule.decreaseListenerCount();
     }
     var name = convertEventEnum(eventName);
     AirshipNotificationEmitter.removeListener(name, listener);
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -352,10 +352,13 @@ class UrbanAirship {
    *
    * @param {string} eventName The event name. Either notificationResponse, pushReceived,
    * register, deepLink, or notificationOptInStatus.
-   * @param {Function} listener The event listner.
+   * @param {Function} listener The event listener.
    * @return {EmitterSubscription} An emitter subscription.
    */
   static addListener(eventName: UAEventName, listener: Function): EmitterSubscription {
+    if (Platform.OS === 'android') {
+      UrbanAirshipModule.addListener();
+    }
     var name = convertEventEnum(eventName);
     return AirshipNotificationEmitter.addListener(name, listener);
   }
@@ -365,9 +368,12 @@ class UrbanAirship {
    *
    * @param {string} eventName The event name. Either notificationResponse, pushReceived,
    * register, deepLink, or notificationOptInStatus.
-   * @param {Function} listener The event listner.
+   * @param {Function} listener The event listener.
    */
   static removeListener(eventName: AirshipEventName, listener: Function) {
+    if (Platform.OS === 'android') {
+      UrbanAirshipModule.removeListener();
+    }
     var name = convertEventEnum(eventName);
     AirshipNotificationEmitter.removeListener(name, listener);
   }
--- a/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
+++ b/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
@@ -36,11 +36,12 @@ class EventEmitter {
     }
 
     /**
-     * Adds an event listener.
+     * Increases the event listener count.
+     * Invokes any pending events.
      *
      * @param context The application context.
      */
-    void addListener(Context context) {
+    void increaseListenerCount(Context context) {
         listenerCount++;
         if (listenerCount > 0 && pendingEvents.size() > 0) {
             for (Event event : pendingEvents) {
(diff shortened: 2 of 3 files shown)


js/UrbanAirship.js
```diff
@@ -365,9 +368,12 @@ class UrbanAirship {
    *
    * @param {string} eventName The event name. Either notificationResponse, pushReceived,
    * register, deepLink, or notificationOptInStatus.
-   * @param {Function} listener The event listner.
+   * @param {Function} listener The event listener.
    */
   static removeListener(eventName: AirshipEventName, listener: Function) {
+    if (Platform.OS === 'android') {
+      UrbanAirshipModule.decreaseListenerCount();
```
rlepinski: This may cause some issues if the subscription was cancelled from the return type of addListener of if removeListener is called multiple times with the same values. We should probably extend the class like react does and decrease the count in removeSubscription - https://github.com/facebook/react-native/blob/a93b7a2da0849f15708afef1ff99022d8cc55b14/Libraries/EventEmitter/NativeEventEmitter.js#L53

bevkoski: I agree that what you suggest is a more correct approach. But it would mean building a custom `NativeEventEmitter`, since extending the existing one wouldn't do the job. This is because React Native 0.51.0 [introduced some changes](https://github.com/facebook/react-native/commit/2fff445b13da114cc5238ad9064ed2100030d5b9#diff-cb16de13bba13fdbc13d117d51774caf) in the `NativeEventEmitter` class. Before 0.51.0 `_nativeModule.removeListener` was only triggered on iOS. Now, with this new version, it would also be triggered on Android. So this custom event emitter should probably have to extend [`EventEmitter` itself](https://github.com/facebook/react-native/blob/master/Libraries/vendor/emitter/EventEmitter.js) in order to avoid this inconsistency.

Anyhow, in this particular case, where it is used only for `increaseListenerCount` / `decreaseListenerCount`, what I coded shouldn't lead to any major issues. I tested it on our app, where I can reproduce the [issue I reported](https://github.com/urbanairship/react-native-module/issues/58), and I can confirm that this PR fixes it.

rlepinski: I see. So we either have to force react > 0.51 or build our own. The react extension is pretty minor, providing our own is not bad:

```
// @flow
'use strict';

const EventEmitter = require('EventEmitter');
const RCTDeviceEventEmitter = require('RCTDeviceEventEmitter');

const invariant = require('fbjs/lib/invariant');

import type EmitterSubscription from 'EmitterSubscription';

type NativeModule = {
  +addListener: (eventType: string) => void,
  +removeListeners: (count: number) => void,
};

class UrbanAirshipEventEmitter extends EventEmitter {
  _nativeModule: ?NativeModule;

  constructor(nativeModule: ?NativeModule) {
    super(RCTDeviceEventEmitter.sharedSubscriber);
    invariant(nativeModule, 'Native module cannot be null.');
    this._nativeModule = nativeModule;
  }

  addListener(eventType: string, listener: Function, context: ?Object): EmitterSubscription {
    this._nativeModule.addListener(eventType);
    return super.addListener(eventType, listener, context);
  }

  removeAllListeners(eventType: string) {
    invariant(eventType, 'eventType argument is required.');
    const count = this.listeners(eventType).length;
    this._nativeModule.removeListeners(count);
    super.removeAllListeners(eventType);
  }

  removeSubscription(subscription: EmitterSubscription) {
    this._nativeModule.removeListeners(1);
    super.removeSubscription(subscription);
  }
}

module.exports = UrbanAirshipEventEmitter;
```

Then we just need to update the methods you provided to be `removeListeners` and `addListener`

rlepinski:
@bevkoski Could you try out https://github.com/urbanairship/react-native-module/pull/67 before I release it? Want to make sure it fixes your issue.

bevkoski:
@rlepinski, no, it does not work for me. I get errors during the build process of my app.

```
Unable to resolve module `EventEmitter`
from `node_modules/urbanairship-react-native/js/UAEventEmitter.js`:
Module does not exist in the module map
```

If I comment out https://github.com/urbanairship/react-native-module/pull/67/files#diff-f639d01f3ab82eea5f41e009bcfaf469R8, I get:

```
Unable to resolve module `RCTDeviceEventEmitter`
from `node_modules/urbanairship-react-native/js/UAEventEmitter.js`:
Module does not exist in the module map
```

If I then replace https://github.com/urbanairship/react-native-module/pull/67/files#diff-f639d01f3ab82eea5f41e009bcfaf469R9 with `import { DeviceEventEmitter } from 'react-native'`, the build passes.

You cannot simply `require` `EventEmitter` or `RCTDeviceEventEmitter`. You need to import them from `'react-native'`.
The thing is, React Native does not expose `EventEmitter`, like it does `DeviceEventEmitter`: https://github.com/facebook/react-native/blob/26038f50bb003eec3770e2b66d428fbb739f1ce2/Libraries/react-native/react-native-implementation.js#L109.
It only uses it internally from its vendor folder: https://github.com/facebook/react-native/blob/1e8f3b11027fe0a7514b4fc97d0798d3c64bc895/Libraries/vendor/emitter/EventEmitter.js.

rlepinski:
Fixed with https://github.com/urbanairship/react-native-module/pull/67

PR #65, opened 2017-12-22
Release 1.3.1

crow:
objectForKey was being used instead of boolForKey when checking the auto launch flag. Manually tested.

### DIFF
--- a/sample/AirshipSample/package.json
+++ b/sample/AirshipSample/package.json
@@ -1,6 +1,6 @@
 {
   "name": "AirshipSample",
-  "version": "1.2.2",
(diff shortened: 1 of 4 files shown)


sample/AirshipSample/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "AirshipSample",
-  "version": "1.2.2",
+  "version": "1.3.1",
   "private": true,
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start"
```
BrianBatchelder: Do you need to update the version in line 11?

PR #67, opened 2018-01-13
Fix delayed events on Android

rlepinski:
Fixes issue #58 

Sends the addListener and removeListeners call to both iOS and Android so we know if we have a listener.


rlepinski:
@bevkoski Updated this based on your last comment in the other PR.

Now its extending native event listener and calling android versions of the add/remove calls to avoid conflicts in the future.

What react version are you using? I want to make sure it will build for you.

Issue #73, opened 2018-02-12
IN-APP MESSAGE


digitalhobo:
Hi

Should the IN-APP MESSAGE trigger an event when setAutoLaunchDefaultMessageCenter is set to false.

Thanks in advance
:0)

rlepinski:
In-app messages are not related to message center. I understand the confusing as they are both technically a way to message the user in the app, but they are two different features.

digitalhobo:
Thanks for the explanation. 
But if setAutoLaunchDefaultMessageCenter is set to true and you tap the in app notification it opens the message in the message center.

I am right to say that it is the one that pops up at the bottom when the app comes into the foreground.

Thanks you for the quick response.

rlepinski:
Yes, the banner looking things - https://docs.urbanairship.com/whats-new/2015-05-31-in-app-messaging/#main

digitalhobo:
Yes, thats the one, if an IN-APP MESSAGE is included with the message center and setAutoLaunchDefaultMessageCenter is set to false, is it possible to listen to an event.
As I would like to show the message as an overlay.
UrbanAirship.displayMessage(id, true);
rather than goto the message center.

:0)

rlepinski:
So you want an event for when either a push or an in-app message is interacted with that tells the app to display the in-app message?

digitalhobo:
Yes, a push message works as expected.
When we tap a push message, that is associated with a message center message, the event "notificationResponse" is triggered, and the response object contains "_uamid" which seems to be the message center message "id"
which we can then use to open the message in an overlay.

UrbanAirship.displayMessage(response.notification.extras._uamid, true);

rlepinski:
We will introduce a new event in 1.5 that will let you know to launch either the message or message center if setAutoLaunchDefaultMessageCenter is set to false. Then you can use that instead of `_uamid` extra.

digitalhobo:
thanks 👍

rlepinski:
Have a PR up for the work. Going to expose a "showInbox" event that contains the optional message ID. Will only be emitted if setAutoLaunchDefaultMessageCenter is false.

rlepinski:
@digitalhobo Released in 1.5.0

PR #85, opened 2018-03-03
[GH-73] Add show inbox event

rlepinski:
Adds an event to show the inbox if auto launch message center is disabled. Needed to route all inbox events through the application if the app built an react native message center.

Testing:
 - Verified it works on iOS and Android


android/src/main/java/com/urbanairship/reactnative/ReactAutopilot.java
```diff
@@ -102,4 +86,53 @@ public boolean apply(ActionArguments actionArguments) {
...
+                return ActionResult.newEmptyResult();
+            }
+        }
+    }
+
+
+
+
+
+
```
rlepinski: whitespace

crow:
🍔

PR #90, opened 2018-03-12
Queue up push received and registration events

kmagiera:
Previously these types of events would get discarded when launched before the module is initialized and before all the listeners are set in JS (which was the case primarily for background notifications that would trigger app launch)

### DIFF
--- a/ios/UARCTModule/UARCTEventEmitter.m
+++ b/ios/UARCTModule/UARCTEventEmitter.m
@@ -41,27 +41,33 @@ static UARCTEventEmitter *sharedEventEmitter_;
     return self;
 }
 
-- (BOOL)sendEventWithName:(NSString *)eventName body:(id)body {
+- (void)sendEventWithName:(NSString *)eventName body:(id)body {
     if (self.bridge && self.isObserving) {
         [self.bridge enqueueJSCall:@"RCTDeviceEventEmitter"
                             method:@"emit"
                               args:body ? @[eventName, body] : @[eventName]
                         completion:NULL];
 
-        return YES;
+    } else {
+        [self.pendingEvents addObject:@{ @"name": eventName, @"body": body}];
     }
-
-    return NO;
 }
 
 - (void)addListener:(NSString *)eventName {
     self.listenerCount++;
-    if (self.listenerCount > 0) {
-        for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
+    if (self.pendingEvents.count > 0) {
+        NSMutableArray *discardedItems = [NSMutableArray array];
+
+        // capture "count" before we start iterating in case one of the method adds new objects to the queue
+        for (NSUInteger i = 0, count = self.pendingEvents.count; i < count; i++) {
+            NSDictionary *event = [self.pendingEvents objectAtIndex:i];
+            if ([event[@"name"] isEqualToString:eventName]) {
+                [self sendEventWithName:event[@"name"] body:event[@"body"]];
+                [discardedItems addObject:event];
+            }
         }
 
-        [self.pendingEvents removeAllObjects];
+        [self.pendingEvents removeObjectsInArray:discardedItems];
     }
 }
 
@@ -79,9 +85,7 @@ static UARCTEventEmitter *sharedEventEmitter_;
 -(void)deepLinkReceived:(NSString *)deepLink {
     id body = @{ @"deepLink" : deepLink };
 
-    if (![self sendEventWithName:UARCTDeepLinkEventName body:body]) {
-        [self.pendingEvents addObject:@{ @"name": UARCTDeepLinkEventName, @"body": body}];
-    }
+    [self sendEventWithName:UARCTDeepLinkEventName body:body];
 }
 
 #pragma mark -
@@ -96,7 +100,6 @@ static UARCTEventEmitter *sharedEventEmitter_;
 
 -(void)receivedBackgroundNotification:(UANotificationContent *)notificationContent
                     completionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
-
     [self sendEventWithName:UARCTPushReceivedEventName body:[UARCTEventEmitter eventBodyForNotificationContent:notificationContent]];
     completionHandler(UIBackgroundFetchResultNoData);
 }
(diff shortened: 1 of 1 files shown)


rlepinski:
@kmagiera we avoided queuing up push received events because the app may never be ready to receive certain types of background/content available pushes. What does this solve for you in your app?

Rnorback:
@rlepinski Can you explain a bit more about why the app may never be ready to receive certain types of background pushes?

rlepinski:
@kmagiera Does the JS context get initialized even on content-available pushes in the BG? Or does react wait until the app is foregrounding?

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -90,14 +95,22 @@ -(void)deepLinkReceived:(NSString *)deepLink {
...
     completionHandler();
 }
 
 -(void)receivedBackgroundNotification:(UANotificationContent *)notificationContent
                     completionHandler:(void (^)(UIBackgroundFetchResult))completionHandler {
 
-    [self sendEventWithName:UARCTPushReceivedEventName body:[UARCTEventEmitter eventBodyForNotificationContent:notificationContent]];
+    NSDictionary *body = [UARCTEventEmitter eventBodyForNotificationContent:notificationContent];
+
+    if (![self sendEventWithName:UARCTPushReceivedEventName body:body]) {
```
rlepinski: is there any event left that will get dropped on the ground? Might be able to move the add to pending events logic into sendEventWithName:body:

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -55,13 +55,18 @@ - (BOOL)sendEventWithName:(NSString *)eventName body:(id)body {
 }
 
 - (void)addListener:(NSString *)eventName {
-    self.listenerCount++;
-    if (self.listenerCount > 0) {
+   self.listenerCount++;
```
rlepinski: whitespace issue here

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -55,13 +55,18 @@ - (BOOL)sendEventWithName:(NSString *)eventName body:(id)body {
...
+    if (self.pendingEvents.count > 0) {
+        NSMutableArray *discardedItems = [NSMutableArray array];
+
         for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
+            if ([event[@"name"] isEqualToString:eventName] &&
+                [self sendEventWithName:event[@"name"] body:event[@"body"]]) {
```
rlepinski: can this ever return `NO` in this context ?

kmagiera:
Hey @rlepinski when the app is launch in background the JS context will initialize as normal, the only think is that the UI won't be rendered but it is ok to register push listener then and do some stuff in background (like for example send a network request). This is how we use it now with this patch in

rlepinski:
Thanks for verifying. This PR looks fine, minor style issue and a few questions. Could you also fill out https://docs.urbanairship.com/contribution-agreement/ ? Ill probably wait to release this change until I investigate Android to see if I can match functionality https://facebook.github.io/react-native/docs/headless-js-android.html

ios/UARCTModule/UARCTEventEmitter.m
kmagiera: It *should* always pass in this context but just in case it doesn't get delivered (e.g. the condition changes in the future) it is better not to remove it from the queue

ios/UARCTModule/UARCTEventEmitter.m
kmagiera: Makes sense, I can def change that

rlepinski: Be careful here, the other method during `addListener:` might modify the array that you are iterating over. Probably want to make a copy of it in `addListener:` if you do move it into sendEventWIthName

kmagiera:
Hey @rlepinski thanks for the review. I'll address your comments today or tomorrow at latest. I'll be looking into Android too cause I want this to work the same on both platforms, so you can just wait for my other PR

kmagiera:
Thanks again @rlepinski for the review. I updated this PR to respond to your comments. Please take a look when you get a chance. Will also likely be looking into doing the same for Android but in case any changes are needed I'll submit a separate PR

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -41,27 +41,33 @@ - (instancetype)init {
...
-
-    return NO;
 }
 
 - (void)addListener:(NSString *)eventName {
     self.listenerCount++;
-    if (self.listenerCount > 0) {
-        for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
+    if (self.pendingEvents.count > 0) {
```
rlepinski: Pretty sure we can rewrite this block as:
```
  for (id event in [self.pendingEvents copy]) {
        if ([event[@"name"] isEqualToString:eventName]) {
            [self sendEventWithName:event[@"name"] body:event[@"body"]];
            [self.pendingEvents removeObject:event];
        }
    }
```

rlepinski:
One last thing and then its ready to merge

kmagiera:
Hey @rlepinski sorry for not getting back for a while. This PR got buried under tons of other github notifications for me.

Thanks so much for your review. I'll update this PR and fill in the agreement next week.

In the meantime wanted to get back to you as I've been looking into doing the same thing on Android. What I found was that `isCritical` flag being set for most of the notification types was preventing notifications from being delivered in a very similar manner to what's been happening on iOS. So the problem again was that when the app got started in background we would start initializing react context but by the time it happens push would already be consumed and since they don't queue up it would never be delivered. 

So in order to fix that I completely removed the notion of "critical" events. https://github.com/1debit/react-native-module/commit/a332809d Would like to learn a bit more about the reasoning behind adding that flag in the first place. I believe it shouldn't be there because it introduces inconsistencies. As in some events may get delivered and other wont depending on how fast RN context would spin up.

ralepinski:
@kmagiera Yeah, the isCritical flag is the thing you want to remove and make sure the addAndroidEventListener method is checking the name for pending events. 

The isCritical flag was added in attempt to make it more consistent. On Android, channel and push received events can come in without ever initializing the react context. If the app is not cleaned up by the OS, the app could get bombarded by old push received and channel registration events the next time it foregrounds. So the isCritical flag was added to any event that should only happen in the foreground, like push opened and deep links. I am not sure how https://facebook.github.io/react-native/docs/headless-js-android.html works, but removing the isCritical flag for apps using that should make it more consistent so I am still onboard with the changes. 

I will release these changes in a major release since the behavior for those events are changing significantly. We should also add a timestamp and only keep the latest channel registration event. I can make those changes in a different PR though.

rlepinski:
Fixed in https://github.com/urbanairship/react-native-module/pull/118

PR #93, opened 2018-03-14
[CUSTENG-37] Add flag to disable react integration at runtime

rlepinski:
Needed by a user who is trying disable react at runtime for iPad devices.


marc-scig:
🇫🇷

PR #118, opened 2018-05-15
[MOBILE-179] Queue up push received and registration events

crow:
This is cleaned up work from PR#90. 

- drops isCritical in Android
- adds functionality to only emit pending events when a matching listener is added in both iOS and Android

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
+++ b/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
@@ -23,8 +23,12 @@ import java.util.List;
  */
 class EventEmitter {
 
+
+    long listenerCount;
+    List<Event> pendingEvents = new ArrayList<>();
+
     private static EventEmitter sharedInstance = new EventEmitter();
-    private final List<Event> pendingEvents = new ArrayList<>();
+
     private ReactInstanceManager reactInstanceManager;
     private boolean isEnabled;
 
@@ -36,7 +40,6 @@ class EventEmitter {
     static EventEmitter shared() {
         return sharedInstance;
     }
-
 
     /**
      * Enables/disables the event emitter.
@@ -74,19 +77,16 @@ class EventEmitter {
             new Handler(Looper.getMainLooper()).post(new Runnable() {
                 @Override
                 public void run() {
-                    sendEvent(applicationContext, event);
+                    if (listenerCount > 0) {
+                        sendEvent(applicationContext, event);
+                    } else {
+                        synchronized (pendingEvents) {
+                            pendingEvents.add(event);
+                        }
+                    }
+
                 }
             });
-
-            return;
-        }
-
-        if (!isEnabled) {
-            if (event.isCritical()) {
-                synchronized (pendingEvents) {
-                    pendingEvents.add(event);
-                }
-            }
 
             return;
         }
@@ -100,7 +100,7 @@ class EventEmitter {
         ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
         if (reactContext != null && reactContext.hasActiveCatalystInstance()) {
             emit(reactContext, event.getName(), event.getBody());
-        } else if (event.isCritical() || reactInstanceManager.hasStartedCreatingInitialContext()) {
+        } else if (reactInstanceManager.hasStartedCreatingInitialContext()) {
             reactInstanceManager.addReactInstanceEventListener(new ReactInstanceManager.ReactInstanceEventListener() {
                 public void onReactContextInitialized(ReactContext reactContext) {
                     emit(reactContext, event.getName(), event.getBody());
--- a/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
+++ b/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
@@ -29,7 +29,7 @@ class EventEmitter {
 
 
     private long listenerCount;
-    private final List<Event> pendingEvents = new ArrayList<>();
+    private List<Event> pendingEvents = new ArrayList<>();
     private Set<String> knownListeners = new HashSet<>();
 
     private static EventEmitter sharedInstance = new EventEmitter();
@@ -121,17 +121,18 @@ class EventEmitter {
      */
     void addAndroidListener(ReactContext reactContext, String eventName) {
         synchronized (knownListeners) {
-            List<Event> pending = pendingEvents;
+            List<Event> pending = new ArrayList<>();
+            pending.addAll(pendingEvents);
 
-            for (Event event : pending) {
-                if (event.getName().equals(eventName)) {
+            for (Event event : pendingEvents) {
+                if (event.equals(eventName)) {
                     sendEvent(reactContext, event);
-                    pendingEvents.remove(event);
+                    pending.remove(event);
                 }
             }
 
-            setEnabled(reactContext, true);
-
+            pendingEvents.clear();
+            pendingEvents.addAll(pending);
 
             listenerCount++;
             knownListeners.add(eventName);
@@ -150,7 +151,7 @@ class EventEmitter {
 
             if (listenerCount == 0) {
                 setEnabled(reactContext, false);
-                knownListeners.removeAll(knownListeners);
+                knownListeners.clear();
             }
         }
     }
--- a/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
+++ b/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
@@ -16,7 +16,9 @@ import com.facebook.react.modules.core.RCTNativeAppEventEmitter;
 import com.urbanairship.Logger;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Emits events to listeners in the JS layer.
@@ -26,6 +28,7 @@ class EventEmitter {
 
     long listenerCount;
     List<Event> pendingEvents = new ArrayList<>();
+    Set<String> knownListeners = new HashSet<>();
 
     private static EventEmitter sharedInstance = new EventEmitter();
 
@@ -77,7 +80,7 @@ class EventEmitter {
             new Handler(Looper.getMainLooper()).post(new Runnable() {
                 @Override
                 public void run() {
-                    if (listenerCount > 0) {
+                    if (listenerCount > 0 && knownListeners.contains(event)) {
                         sendEvent(applicationContext, event);
                     } else {
                         synchronized (pendingEvents) {
@@ -109,6 +112,23 @@ class EventEmitter {
             });
         }
     }
+
+    /**
+     * Helper method to add known listeners.
+     *
+     * @param eventName The event name identifying the listener.
+     */
+    void addKnownListener(String eventName) {
+        knownListeners.add(eventName);
+    }
+
+    /**
+     * Helper method to remove all known listeners.
+     */
+    void removeKnownListeners() {
+        knownListeners.removeAll(knownListeners);
+    }
+
 
     /**
      * Helper method to emit data.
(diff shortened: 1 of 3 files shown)


rlepinski:
So @BrianBatchelder pointed out an issue. If you add a single listener, lets say "foo", pending events will stop accumulating for other event types like "bar". A proposed fix would be to have a list of known listeners and check that list when decided to add a pending event.

android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -139,13 +154,15 @@ public void addAndroidListener(String eventName) {
     @ReactMethod
     public void removeAndroidListeners(int count) {
         Logger.info("UrbanAirshipReactModule - Event listeners removed: " + count);
-        listenerCount -= count;
-        if (listenerCount < 0) {
-            listenerCount = 0;
+
+        EventEmitter.shared().listenerCount -= count;
+        if (EventEmitter.shared().listenerCount < 0) {
```
marc-scig: not a critical thing, but you could also write this as:

```
long currentCount = EventEmitter.shared().listenerCount;
EventEmitter.shared().listenerCount = max(0, currentCount - count);
```

Or something. Sometimes it can be easier to think in terms of clipping value ranges, and if it were behind a getter/setter you could encapsulate the clipping there. Up to you though.

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -36,37 +40,44 @@ - (instancetype)init {
...
     self.listenerCount++;
-    if (self.listenerCount > 0) {
-        for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
-        }
 
-        [self.pendingEvents removeAllObjects];
+    [self.knownListeners addObject:eventName];
+
+    if (self.pendingEvents.count > 0) {
```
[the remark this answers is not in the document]
crow: 👍

android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -127,8 +127,20 @@ public String getName() {
...
-        listenerCount++;
-        EventEmitter.shared().setEnabled(getReactApplicationContext(), true);
+        EventEmitter.shared().listenerCount++;
+        EventEmitter.shared().addKnownListener(eventName);
+
+        List<Event> pending = EventEmitter.shared().pendingEvents;
+
+        synchronized (EventEmitter.shared().pendingEvents) {
+            for (Event event : pending) {
+                if (event.getName() == eventName) {
```
rlepinski: need to do a string equals check here

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -36,37 +40,51 @@ - (instancetype)init {
...
 
 - (void)addListener:(NSString *)eventName {
     self.listenerCount++;
-    if (self.listenerCount > 0) {
-        for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
-        }
 
-        [self.pendingEvents removeAllObjects];
+    @synchronized(self.knownListeners) {
```
BrianBatchelder: I think you need to add the eventName to the listeners after you flush the pending events. As it is, if an event arrives before you are done flushing, that event might be passed to the listener out of order.

android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -127,8 +127,20 @@ public String getName() {
     @ReactMethod
     public void addAndroidListener(String eventName) {
         Logger.info("UrbanAirshipReactModule - Event listener added: " + eventName);
-        listenerCount++;
-        EventEmitter.shared().setEnabled(getReactApplicationContext(), true);
+        EventEmitter.shared().listenerCount++;
+        EventEmitter.shared().addKnownListener(eventName);
```
BrianBatchelder: I think you need to add the eventName to the listeners after you flush the pending events. As it is, if an event arrives before you are done flushing, that event might be passed to the listener out of order.

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -36,37 +40,50 @@ - (instancetype)init {
...
+        @synchronized(self.pendingEvents) {
+            [self.pendingEvents addObject:@{ UARCTEventNameKey: eventName, UARCTEventBodyKey: body}];
+        }
     }
-
-    return NO;
 }
 
 - (void)addListener:(NSString *)eventName {
     self.listenerCount++;
```
marc-scig: I have to admit I'm a little confused by the synchronization going on here, but that's probably because it's not clear to me what kind of multithreading environment this stuff is living in. But for instance, take our methods for adding and removing listeners. If it's necessary to synchronize on knownListeners in those methods, presumably because they can be called from different threads, should mutating self.listenerCount also be synchronized?

ios/UARCTModule/UARCTEventEmitter.m
```diff
@@ -36,37 +40,50 @@ - (instancetype)init {
...
-        for (NSDictionary *event in self.pendingEvents) {
-            [self sendEventWithName:event[@"name"] body:event[@"body"]];
+
+    @synchronized(self.pendingEvents) {
+        for (id event in [self.pendingEvents copy]) {
+            if ([event[UARCTEventNameKey] isEqualToString:eventName]) {
+                [self sendEventWithName:event[UARCTEventNameKey] body:event[UARCTEventBodyKey]];
+                [self.pendingEvents removeObject:event];
+            }
         }
```
BrianBatchelder: Shoot. Now an event arriving during this loop will be added to `pendingEvents`, and never sent to the app. @marc-scig might be referring to the same thing. I think the entire `sendEventWithName:` and `addListener:` methods need to be synchronized. Don't do anything in either if the other one is executing. Maybe that can be done with `@synchronized(self.knownListeners)` at the start of each method?

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -16,15 +16,22 @@
...
 
 /**
  * Emits events to listeners in the JS layer.
  */
 class EventEmitter {
 
+
+    long listenerCount;
+    List<Event> pendingEvents = new ArrayList<>();
+    Set<String> knownListeners = new HashSet<>();
```
rlepinski: make this private final

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -16,15 +16,22 @@
...
+import java.util.Set;
 
 /**
  * Emits events to listeners in the JS layer.
  */
 class EventEmitter {
 
+
+    long listenerCount;
+    List<Event> pendingEvents = new ArrayList<>();
```
rlepinski: private final

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -16,15 +16,22 @@
...
 import java.util.List;
+import java.util.Set;
 
 /**
  * Emits events to listeners in the JS layer.
  */
 class EventEmitter {
 
+
+    long listenerCount;
```
rlepinski: private

android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -127,8 +128,24 @@ public String getName() {
     @ReactMethod
     public void addAndroidListener(String eventName) {
         Logger.info("UrbanAirshipReactModule - Event listener added: " + eventName);
-        listenerCount++;
-        EventEmitter.shared().setEnabled(getReactApplicationContext(), true);
+
+        synchronized (EventEmitter.shared().knownListeners) {
```
rlepinski: this code should live in EventEmitter

rlepinski: I would just call through to event emitter: `EventEmitter.shared().addListener(eventName)`

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
rlepinski: private

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -110,6 +112,23 @@ public void onReactContextInitialized(ReactContext reactContext) {
         }
     }
 
+    /**
+     * Helper method to add known listeners.
+     *
+     * @param eventName The event name identifying the listener.
+     */
+    void addKnownListener(String eventName) {
```
rlepinski: make this all internal

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -114,22 +115,46 @@ public void onReactContextInitialized(ReactContext reactContext) {
...
+     * Called when a new listener is added for a specified event name.
      *
-     * @param eventName The event name identifying the listener.
+     * @param eventName The event name.
      */
-    void addKnownListener(String eventName) {
-        knownListeners.add(eventName);
+    void addAndroidListener(ReactContext reactContext, String eventName) {
+        synchronized (knownListeners) {
+            List<Event> pending = pendingEvents;
```
rlepinski: pretty sure this will cause an error. you need to copy the list, not just point a different variable to it

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -110,6 +114,48 @@ public void onReactContextInitialized(ReactContext reactContext) {
         }
     }
 
+    /**
+     * Called when a new listener is added for a specified event name.
+     *
+     * @param eventName The event name.
+     */
+    void addAndroidListener(ReactContext reactContext, String eventName) {
+        synchronized (knownListeners) {
```
rlepinski:
```
void addAndroidListener(ReactContext reactContext, String eventName) {
    synchronized (knownListeners) {
        for (Event event : new ArrayList<>(pendingEvents)) {
            if (event.equals(eventName)) {
                sendEvent(reactContext, event);
                pendingEvents.remove(event);
            }
        }
        listenerCount++;
        knownListeners.add(eventName);
    }
}
```

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -110,6 +114,48 @@ public void onReactContextInitialized(ReactContext reactContext) {
...
+     *
+     * @param count The count of listeners.
+     */
+    void removeAndroidListeners(ReactContext reactContext, int count) {
+        synchronized (knownListeners) {
+            long currentCount = listenerCount;
+            listenerCount = max(0, currentCount - count);
+
+            if (listenerCount == 0) {
+                setEnabled(reactContext, false);
```
rlepinski: do we still need the enabled flag?

rlepinski: The code in setEnabled could cause an Exception. I dont think its needed anymore so you might be able to just delete it.

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
crow: fixed this, it just didn't change the line you commented on

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -110,6 +91,40 @@ public void onReactContextInitialized(ReactContext reactContext) {
...
 
+    /**
+     * Called when a new listener is added for a specified event name.
+     *
+     * @param eventName The event name.
+     */
+    void addAndroidListener(ReactContext reactContext, String eventName) {
+        synchronized (knownListeners) {
+            for (Event event : new ArrayList<>(pendingEvents)) {
+                if (event.equals(eventName)) {
```
rlepinski: event.getName().equals(eventName)

rlepinski:
👍

PR #125, opened 2018-06-13
[MOBILE-48] fix repo ordering, update gradle build tools

crow:
This fixes an error I was coming across when getting CI working. The ordering of the repositories matters here, for whatever reason. Also updated gradle build tools while I was at it.


rlepinski:
👍

PR #135, opened 2018-07-30
2.0 fixes

marc-scig:
### What do these changes do?
This reverts the change to the firebase core dependency, and also restructures the sample app to have the same package structure as our main android sample, so that it's compatible with the google-services.json and easier to test. Feel free to suggest otherwise, it just seemed like a good thing to do.

I accidentally put in the wrong version number for firebase core and in the rush to get it out didn't realize that it broke the build.

Made sure the android sample builds and runs correctly with this change. This may still require extra work to get around the cryptic "window error" that bdb was reporting this morning, and which appears to be tooling related.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,5 +13,5 @@ android {
 dependencies {
     compile 'com.facebook.react:react-native:[0.40,)'
     compile 'com.urbanairship.android:urbanairship-fcm:9.4.1'
-    compile 'com.google.firebase:firebase-core:16.0.0'
+    compile 'com.google.firebase:firebase-core:16.0.1'
 }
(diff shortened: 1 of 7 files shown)


android/build.gradle
```diff
@@ -13,5 +13,5 @@ android {
 dependencies {
     compile 'com.facebook.react:react-native:[0.40,)'
     compile 'com.urbanairship.android:urbanairship-fcm:9.4.1'
-    compile 'com.google.firebase:firebase-core:17.0.0'
+    compile 'com.google.firebase:firebase-core:16.0.0'
```
rlepinski: 16.0.1 ?

marc-scig: sure.

rlepinski: https://firebase.google.com/docs/android/setup#available_libraries

rlepinski:
Change log + version bump?

marc-scig:
@rlepinski lets do that in a release ticket

rlepinski:
👍

PR #139, opened 2018-08-03
Address security vulnerabilities.

BrianBatchelder:
### What do these changes do?
This PR updates libraries with known security vulnerabilities to ones without known vulnerabilities. A dependency on `npm` was removed because it isn't necessary, and was exposing the plugin to some security vulnerabilities of its own.

Github detected, and `npm audit` confirmed, that our plugin contained libraries with known security vulnerabilities.

Local `npm audit`. Will monitor GitHub after PR merge.


package.json
```diff
@@ -30,7 +30,5 @@
   },
   "license": "Apache-2.0",
   "typings": "./js/index.d.ts",
-  "dependencies": {
-    "npm": "^5.8.0"
```
marc-scig: do we still want the npm dependency?

BrianBatchelder: It isn't necessary, and was exposing us to some security vulnerabilities of its own.

rlepinski: No idea why it was in there.

rlepinski:
👍

Issue #128, opened 2018-07-13
iOS 'React/RCTEventEmitter.h' file not found


wpconsulate:
I detached an Expo project and using expo-kit to develop and I followed the instructions here https://docs.urbanairship.com/platform/react-native/ and installed the SDK. While building I'm getting this error http://prntscr.com/k61d3g. I tested the App working fine just before installing the UrbanAirship SDK, So, something for sure wrong with UrbanAirship SDK.

# Preliminary Info

### What Urban Airship dependencies are you using?
"urbanairship-react-native": "^1.6.2"
I have put in podfile 
`pod 'UrbanAirship-iOS-SDK', '~>9.0'`
and run the "pod install"

### What are the versions of any relevant development tools you are using?
My Package.json file looks like this

> {
>   "main": "node_modules/expo/AppEntry.js",
>   "private": true,
>   "dependencies": {
>     "expo": "^28.0.0",
>     "jsonq": "^1.1.0",
>     "moment": "^2.22.2",
>     "react": "16.3.1",
>     "react-native": "https://github.com/expo/react-native/archive/sdk-27.0.0.tar.gz",
>     "react-native-datepicker": "^1.7.2",
>     "react-native-device-info": "^0.21.5",
>     "react-native-elements": "^0.19.1",
>     "react-native-google-analytics": "^1.3.2",
>     "react-native-keyboard-aware-scroll-view": "^0.5.0",
>     "react-native-masked-text": "^1.7.2",
>     "react-native-picker-select": "^3.1.3",
>     "react-navigation": "^2.5.5",
>     "urbanairship-react-native": "^1.6.2"
>   }
> }

I tried building the react.xcodeproj Project itself and also tried building the Pods.xcodeproj project as well but no help there. Any suggestions would be appriciated.

### What unexpected behavior are you seeing?
iOS project build is failing

### What is the expected behavior?
iOS project should build as normal

### What are the steps to reproduce the unexpected behavior?
create a react native project with above packagaes and try installing the UrbanTitanium it'll break the build

### Do you have logging for the issue?
please refer to screenshot: http://prntscr.com/k61jxz

rlepinski:
Duplicate of https://github.com/urbanairship/react-native-module/issues/109

Take a look at this issue - https://github.com/facebook/react-native/issues/13371. You have to serialize the build order on react projects or react can fail to build.

wpconsulate:
[...]

PR #147, opened 2018-08-31
add react to header search path

davidgovea:
### What do these changes do?
This PR adds another path to the "Header Search Paths" of the xcode project.

We are using detached ExpoKit, and experienced the same issue as #128 

The [solution listed there](https://github.com/urbanairship/react-native-module/issues/128#issuecomment-405014395) worked for us, but the instructed change is inside the `node_modules` folder, so it does not get tracked by git, and is not a full fix.

This change allows our build to succeed after a fresh installation of node dependencies.

1. Create a new app with CRNA
1. Eject with ExpoKit
1. Install this library - observe issue #128 
1. Install forked version
1. Observe build succeeds

ℹ If applicable, please provide screenshots here.

Looks like there are already some maybe-used search paths already (e.g. Carthage), so adding another probably won't be a big deal.

Thanks UA team!
-David


BrianBatchelder:
Thanks for the PR, David. We'll take a look at it as soon.

rlepinski:
This looks fine to me. 👍

Issue #148, opened 2018-09-02
[Android] Emitted push event are not being handled on app launch


AntiHate:
We are facing exactly the same issue in v2.0.1 as described in #58 but with push notification opened. We are using react-navigation with redux so its takes some time to initialize the listener.

We fixed this by making some changes in EventEmitter.java. 

```
ReactContext reactContext = reactInstanceManager.getCurrentReactContext();
        if (reactContext != null && reactContext.hasActiveCatalystInstance()) {
//HERE --> This execute when app launched after soft exit (hardware back button exit)
            synchronized (knownListeners) {
                if (knownListeners.contains(event.getName())) {
                    emit(reactContext, event.getName(), event.getBody());
                } else if(event.getName() == "com.urbanairship.notification_response") {
                    pendingEvents.add(event);
                }
            }
        } else if (reactInstanceManager.hasStartedCreatingInitialContext()) {
            reactInstanceManager.addReactInstanceEventListener(new ReactInstanceManager.ReactInstanceEventListener() {
                public void onReactContextInitialized(ReactContext reactContext) {
                    emit(reactContext, event.getName(), event.getBody());
                    reactInstanceManager.removeReactInstanceEventListener(this);
                }
            });
        } else {
//HERE --> This executes when app launched after killed (cold launch)
            if(event.getName() == "com.urbanairship.notification_response") {
                pendingEvents.add(event);
            }            
        }
```

I don't know much about native side but this block doesn't execute when app is opened after exit (soft or killed).

```
        // Force the call to be on the main thread
        if (Looper.getMainLooper() != Looper.myLooper()) {
            final Context applicationContext = context.getApplicationContext();
            new Handler(Looper.getMainLooper()).post(new Runnable() {
                @Override
                public void run() {
                    synchronized (knownListeners) {
                        if (knownListeners.contains(event.getName())) {
                            sendEvent(applicationContext, event);
                        } else {
                            pendingEvents.add(event);
                        }
                    }
                }
            });

            return;
        }
```

Any suggestions?

rlepinski:
I think I see the problem. It looks like we only add to pending if sendEvents is called off the main thread. Ill have a PR in a bit to fix it.

rlepinski:
@AntiHate I am working on the branch https://github.com/urbanairship/react-native-module/tree/GH-148, I just need to test it then I will post a PR.

AntiHate:
Thanks @rlepinski for acknowledging the issue.

[...]

PR #150, opened 2018-09-04
GH-148: Fix Android pending events

ralepinski:
### What do these changes do?
Fixes a bug with pending events. The pending event check was only happening when redirecting `sendEvent` to the main looper. If it was called from the main thread it would not check to make sure we have a known listener for the event. I also went ahead and did some cleanup:
- removed the `isCritical` method from events (no longer used)
- simplify the reactContext lookup by having our module attach it to the event manager. No more casting :)
- instead of doing main looper checks, just post to the main thread

Fixes pending events on Android. 

Ran the sample app.

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
+++ b/android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
@@ -27,34 +27,21 @@ import static java.lang.Math.max;
  */
 class EventEmitter {
 
-    private static EventEmitter sharedInstance;
+    private static EventEmitter sharedInstance = new EventEmitter();
 
     private final List<Event> pendingEvents = new ArrayList<>();
     private final Set<String> knownListeners = new HashSet<>();
     private final Handler mainHandler = new Handler(Looper.getMainLooper());
-    private final Context context;
 
     private long listenerCount;
     private ReactContext reactContext;
-
-    private EventEmitter(Context context) {
-        this.context = context.getApplicationContext();
-    }
 
     /**
      * Returns the shared {@link EventEmitter} instance.
      *
      * @return The shared {@link EventEmitter} instance.
      */
-    static EventEmitter shared(Context context) {
-        if (sharedInstance == null) {
-            synchronized (EventEmitter.class) {
-                if (sharedInstance == null) {
-                    sharedInstance = new EventEmitter(context);
-                }
-            }
-        }
-
+    static EventEmitter shared() {
         return sharedInstance;
     }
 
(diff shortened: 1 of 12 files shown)


android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -46,110 +45,110 @@ static EventEmitter shared() {
         return sharedInstance;
     }
 
+    /**
+     * Attaches the react context.
+     *
+     * @param reactContext The react context.
+     */
+    void attachReactContext(final ReactContext reactContext) {
```
ralepinski: This is called in our module when the context is initialized.

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -46,110 +45,110 @@ static EventEmitter shared() {
...
+        mainHandler.post(new Runnable() {
```
ralepinski: Do it on the main thread so we dont have to worry about locking

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -46,110 +45,110 @@ static EventEmitter shared() {
...
-                @Override
-                public void run() {
-                    synchronized (knownListeners) {
-                        if (knownListeners.contains(event.getName())) {
-                            sendEvent(applicationContext, event);
-                        } else {
-                            pendingEvents.add(event);
-                        }
+    void sendEvent(final Event event) {
+        mainHandler.post(new Runnable() {
```
ralepinski: Just post on main thread instead of doing the looper check

android/src/main/java/com/urbanairship/reactnative/EventEmitter.java
```diff
@@ -46,110 +45,110 @@ static EventEmitter shared() {
...
+     * Attempts to send pending events.
      */
-    private void emit(ReactContext reactContext, String eventName, Object eventBody) {
-        reactContext.getJSModule(RCTNativeAppEventEmitter.class).emit(eventName, eventBody);
+    @MainThread
+    private void sendPendingEvents() {
+        synchronized (knownListeners) {
+            for (Event event : new ArrayList<>(pendingEvents)) {
+                if (knownListeners.contains(event.getName())) {
+                    pendingEvents.remove(event);
```
marc-scig: I'm a little confused by this. We're removing the the pending event here, and then in sendEvent we're adding it back if there are no valid listeners or emitting failed? Did I get that right?

rlepinski: yeah, ill add a comment to make that clear

PR #152, opened 2018-09-07
[MOBILE-297] provide a way to enable notifications and get the auth result

crow:
### What do these changes do?
These changes provide an interface for the newly added enableUserPushNotifications function in UAPush. This allows users to enable push with a promise that returns the auth result.

Users want to be able to enable push and receive the authorization result back immediately. 

Manual testing in the sample.

### DIFF
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -154,6 +154,16 @@ class UrbanAirship {
   }
 
   /**
+   * Enables user notifications.
+   *
+   * @return {Promise.<boolean>} A promise that returns true if enablement was authorized
+   * or false if enablement was rejected
+   */
+  static enableUserPushNotifications(): Promise<boolean> {
+    return UrbanAirshipModule.enableUserPushNotifications();
+  }
+
+  /**
    * Checks if app notifications are enabled or not. Its possible to have `userNotificationsEnabled`
    * but app notifications being disabled if the user opted out of notifications.
    *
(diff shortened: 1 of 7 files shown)


crow:
Failing with - /Users/vagrant/git/ios/UARCTModule/UrbanAirshipReactModule.m:71:22: error: no visible @interface for 'UAPush' declares the selector 'enableUserPushNotifications:'. This will go away when the enableUserPushNotifications feature is released in the SDK.

marc-scig:
🐤

js/UrbanAirship.js
```diff
@@ -153,6 +153,16 @@ class UrbanAirship {
     return UrbanAirshipModule.isUserNotificationsEnabled();
   }
 
+  /**
+   * Enables user notifications.
+   *
+   * @return {Promise.<boolean>} A promise that returns true if enablement was authorized
+   * or an false if enablement was rejected
```
BrianBatchelder: or ~an~ false

js/index.d.ts
```diff
@@ -110,6 +110,14 @@ export class UrbanAirship {
      */
     static isUserNotificationsEnabled(): Promise<boolean>;
 
+    /**
+     * Enables user push notifications.
+     *
+     * @return {Promise.<boolean>} A promise that returns true if enablement was authorized
+     * or an false if enablement was rejected.
```
BrianBatchelder: or ~an~ false

rlepinski:
@crow You need to add the Android version as well. It can just enable notifications and return true.

Issue #23, opened 2017-07-06
"notificationResponse" not firing when app is opened from a notification?


Zyphxion:
Maybe I'm going about this the wrong way, but after reading [this migration guide](https://github.com/urbanairship/react-native-module/blob/master/ReactNativeUAMigration.md) it looks like "notificationResponse" should have the same functionality as "launchedFromNotification," however when launching my app from a notification nothing is happening.  All I have in the event is a simple alert, set up the exact same way as the sample project, inside the componentWillMount lifecycle step.

Could there be something I'm missing?  My goal is to have a user open the app via a notification and upon entering the application see an alert that has a message and a few other things from the "extras" field.  I even tried including the message and extra data in the "in_app" attribute in the push object being sent to UA from my server, no dice.

Thanks in advance.

rlepinski:
@Zyphxion what platform OS version? Can you reproduce your issue with the sample?

rlepinski:
Closing for now. If you are still having an issue please reopen with the requested info.

surendharreddy:
@ralepinski this happens for me. I can reproduce it on my side after migrating from `react-native-ua`. 

Implemented it same as @Zyphxion. When launching the app from the notification I don't see my alerts getting triggered. 

I'm using `"urbanairship-react-native": "^2.0.1"` with `"react-native": "0.55.4"`.

BrianBatchelder:
Thanks for the report, @Zyphxion . We will take a look at this. In the meantime, we released version 2.0.2 yesterday. I’m not sure it will address your issue, but you might give it a try.

surendharreddy:
Thanks for reopening this @BrianBatchelder. Look forward to the fix. Let me know if you need any additional details, I would be happy to provide them.

BrianBatchelder:
@surendharreddy: could you try version 2.0.2 and let me know if your issue still exists? I meant to ask you in my previous comment, but copied and pasted the wron user name. Thanks.

surendharreddy:
@BrianBatchelder I tried using version 2.0.2 and the issue still exits. 😞

BrianBatchelder:
@surendharreddy: the migration guide says: 

> UrbanAirship.addListener(eventName: string, listener: Function) requires specifying the event name when adding the `listener. 

Are you calling `addListener` for the `notificationResponse` event?

surendharreddy:
[...]

PR #153, opened 2018-09-13
[GITHUB-23] fix listener addition in UARCTEventEmitter

crow:
### What do these changes do?
These changes fix the issue described in issue 23:
https://github.com/urbanairship/react-native-module/issues/23

More specifically, this changes the location of the known listener addition to before the pending events are sent in the addListener method. This matches the current Android functionality

These changes are necessary to allow the notification response to result in a notification to the js-bridge.

Ran the sample with and without change.


rlepinski:
👍

PR #166, opened 2018-11-20
Release 2.1.3

rlepinski:


rlepinski:
Caret operator looks like it picks up minor and patch https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

PR #171, opened 2018-12-05
[MOBILE-405] Fix docs

BrianBatchelder:
### What do these changes do?
Fix docs problem by removing `@param` for non-existent param.

Docs need to be accurate.

Built docs locally.


rlepinski:
👍

crow:
🍔

PR #179, opened 2019-05-14
[MOBILE-592] add support for custom notification categories

marc-scig:
This adds the ability to define custom notification categories by way of plist/xml files. I followed the approach used here https://github.com/urbanairship/urbanairship-cordova/pull/228

The main difference between cordova and react here is that there's no config.xml, and instead you have add files directly to an Xcode project or save them in res/xml. Otherwise it's the same pattern. Dev supplies `ua_custom_notification_buttons.xml` and/or `UACustomNotificationCategories.plist`, the plugin looks for these files after takeOff, and if present, uses them to load custom categories or button groups.

Tested manually. Docs PR can be found here: https://github.com/urbanairship/extdocs/pull/2833

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactAutopilot.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactAutopilot.java
@@ -2,11 +2,14 @@
 
 package com.urbanairship.reactnative;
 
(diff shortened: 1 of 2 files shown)


rlepinski:
👍

PR #180, opened 2019-05-29
MOBILE-642 Update react plugin to iOS SDK 11.0

marc-scig:
There are two aspects of SDK 11.0 that required changes to the plugin. One was to get rid of old references to `[UAirship location]`. Instead of creating a whole separate plugin, this approach uses `[UAirship shared].locationProviderDelegate`, which calls through to the module (or a custom provider delegate) if present and no-ops otherwise. Since Airship dependencies are added using cocoapods, all a user has to do to enable the module is to add the AirshipLocationKit reference to their pod file. There will be other PRs in a bit that update the docs accordingly.

The other thing that needed work was methods having to do with display message center messages in overlays. We got rid of the old overlay controller in 11.0 (we hardly knew ye) and replaced it with a bridged HTML IAM. An unfortunate and unintended consequence of this is that it makes it hard-to-impossible to dismiss such a message from JS via the plugin itself, rather than the injected native bridge, and our plugin historically has a "close" method for these.

After some discussion, I picked what I thought was the least terrible short term approach, which is to set a custom adapter factory and use KVC and selectors to access the HTML view controller. It's not ideal, but it will work as a stopgap while we figure out how to do this in a cleaner way, which will almost certainly require some kind of SDK update. React users won't need to worry about this, and from their perspective it will just keep working as usual.

### DIFF
--- a/ios/UARCTModule/UrbanAirshipReactModule.m
+++ b/ios/UARCTModule/UrbanAirshipReactModule.m
@@ -136,23 +136,23 @@ RCT_REMAP_METHOD(associateIdentifier,
 }
 
 RCT_EXPORT_METHOD(setLocationEnabled:(BOOL)enabled) {
-    [UAirship location].locationUpdatesEnabled = enabled;
+    [UAirship shared].locationProviderDelegate.locationUpdatesEnabled = enabled;
 }
 
 RCT_REMAP_METHOD(isLocationEnabled,
                  isLocationEnabled_resolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject) {
-    resolve(@([UAirship location].isLocationUpdatesEnabled));
+    resolve(@([UAirship shared].locationProviderDelegate.isLocationUpdatesEnabled));
 }
 
 RCT_REMAP_METHOD(isBackgroundLocationAllowed,
                  isBackgroundLocationAllowed_resolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject) {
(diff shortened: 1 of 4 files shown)


ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -359,7 +361,7 @@ - (RCTBridge *)bridge {
 
 RCT_REMAP_METHOD(displayMessage,
                  messageId:(NSString *)messageId
-                 overlay:(BOOL *)overlay
```
marc-scig: note: these methods previously had BOOL pointers in the signature, which I have to assume was a mistake.

ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -528,4 +530,38 @@ - (RCTBridge *)bridge {
         reject(UARCTStatusUnavailable, @"Only available on iOS 10+", error);
     }
 }
+
+#pragma mark -
+#pragma mark Helper methods
+
+- (void)displayOverlayMessage:(UAInboxMessage *)message {
+    if (!self.factoryBlockAssigned) {
```
BrianBatchelder: Just to verify I understand this, the code will set the factory block regardless if any other code has set one. Not sure what other code would do that...

marc-scig: Right. It's an unlikely scenario, since the app developer is going to be writing some flavor of JS instead of native code. It's technically possible that some random third party plugin could be messing around with Airship APIs but I think that's well within the boundary of unsupported integration.

PR #182, opened 2019-05-31
MOBILE-651 Update android SDK in react native to 10.0

marc-scig:
These are changes to the react native plugin to bring it up to date with our latest android SDK. In particular:

* No more AirshipReceiver. SDK events having to do with things like message center, push and registration events now use the new listener interfaces
* CustomLandingPageActivity has been removed since its superclass is gone
* Landing page overlays have been replaced with HTML in-app messages, so they have to be displayed and dismissed in a way that works with IAA

NotificationFactory was deprecated in SDK 10.0, but I've deferred dealing with changes regarding that class to the notification config PR, since it involves writing a custom NotificationProvider.


android/src/main/java/com/urbanairship/reactnative/ReactAutopilot.java
```diff
@@ -38,38 +40,101 @@
...
+            }
+
+            @Override
+            public void onNotificationBackgroundAction(@NonNull NotificationInfo notificationInfo, @NonNull NotificationActionButtonInfo notificationActionButtonInfo) {
+                Event event = new NotificationResponseEvent(notificationInfo, notificationActionButtonInfo);
+                EventEmitter.shared().sendEvent(event);
+            }
+
+            @Override
+            public void onNotificationDismissed(@NonNull NotificationInfo notificationInfo) {}
```
crow: do we want to keep the stub?

rlepinski: you have to

rlepinski:
👍

PR #183, opened 2019-05-31
[MOBILE-401] add support for loading custom notification channels

marc-scig:
Part of the update to android SDK 10 means adding support for loading custom notification channels. The most straightforward way to do that is to follow the same approach as with notification button groups and provide a mechanism for loading custom notification channels from XML, which is what this does. We will also need to provide basic documentation on how to do this and what the XML structure looks like, which will come in a subsequent extdocs PR.


rlepinski:
👍

PR #184, opened 2019-05-31
[MOBILE-644] migration guide

marc-scig:
There is not much to migrate between the 2.x and 3.0 versions of our react native plugin. However, we should at least mention the change to how location is handled, and link to the appropriate section of the docs. There was no migration guide in the repo previously, so I created one.


rlepinski:
👍

PR #185, opened 2019-05-31
[MOBILE-403] ReactNotificationProvider, notification config support

marc-scig:
In order to provide an interface for configuring notification options, I roughly followed the approach already in place for the cordova plugin. The difference here is mostly in the way react native bridging works, the particulars of how preferences are dealt with at the plugin level, and the fact that  NotifiationFactory has been deprecated in favor of NotificationProvider as of SDK 10.

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
@@ -4,6 +4,7 @@ package com.urbanairship.reactnative;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.support.annotation.Nullable;
 
 import com.urbanairship.UAirship;
 import com.urbanairship.reactnative.events.NotificationOptInEvent;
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
```diff
@@ -21,6 +22,11 @@
 
     private static final String NOTIFICATIONS_OPT_IN_KEY = "NOTIFICATIONS_OPT_IN_KEY";
 
+    private static final String NOTIFICATION_ICON_KEY = "notification_icon";
```
crow: nitpick but might as well follow the convention of making the values all caps

android/src/main/java/com/urbanairship/reactnative/ReactAirshipPreferences.java
```diff
@@ -21,6 +22,11 @@
...
+    private static final String NOTIFICATION_LARGE_ICON_KEY = "notification_large_icon";
+    private static final String NOTIFICATION_ACCENT_COLOR_KEY = "notification_accent_color";
+    private static final String DEFAULT_NOTIFICATION_CHANNEL_ID = "default_notification_channel_id";
```
crow: probably want DEFAULT_NOTIFICATION_CHANNEL_ID**_KEY** for consistency

android/src/main/java/com/urbanairship/reactnative/ReactNotificationProvider.java
```diff
@@ -0,0 +1,102 @@
...
+import android.support.v4.app.NotificationCompat;
+
+import com.urbanairship.AirshipConfigOptions;
+import com.urbanairship.Logger;
+import com.urbanairship.UAirship;
+import com.urbanairship.push.PushMessage;
+import com.urbanairship.push.notifications.AirshipNotificationProvider;
+import com.urbanairship.push.notifications.NotificationArguments;
+
+public class ReactNotificationProvider extends AirshipNotificationProvider {
```
rlepinski: You should update to SDK 10.0.1 and override the getters

marc-scig: Updated.

PR #189, opened 2019-07-02
[MOBILE-509] Fix dev environment

rlepinski:
### What do these changes do?
Make it possible to develop the plugin without copying files over. We are using the approach outlined in https://medium.com/@charpeni/setting-up-an-example-app-for-your-react-native-library-d940c5cf31e4

Makes development easier

Ran app and modified code.


sample/AirshipSample/android/build.gradle
```diff
@@ -22,18 +23,12 @@ buildscript {
 
 allprojects {
     repositories {
-        google()
```
crow: this ordering is important, was this on purpose?

sample/AirshipSample/android/gradlew.bat
```diff
@@ -1,3 +1,19 @@
+@rem
+@rem Copyright 2015 the original author or authors.
```
crow: do we need the correct year for these apache copyrights headers?

rlepinski: this is generated, so no

sample/AirshipSample/android/build.gradle
rlepinski: mavenLocal() normally is first, so its fine. Just needs to be ahead of jcenter()

sample/AirshipSample/package.json
```diff
@@ -1,18 +1,25 @@
 {
   "name": "AirshipSample",
-  "version": "3.0.0",
+  "version": "0.0.1",
```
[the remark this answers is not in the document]
rlepinski: The sample does not need to be versioned. Probably updated when I regenerated the template

sample/AirshipSample/__tests__/App-test.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @format
+ */
+
+import 'react-native';
+import React from 'react';
+import App from '../App';
+
+// Note: test renderer must be required after react-native.
```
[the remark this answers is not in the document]
rlepinski: Another generated file

PR #190, opened 2019-07-03
MOBILE-702: Add support for delayed channel registration

rlepinski:
### What do these changes do?
Adds support for delayed channel registration. To use delayed registration, you have to set `channelCreationDelayEnabled` to `YES`/`true` and then call `enableChannelCreation` on the push manager when you are ready to create a channel. This allows customers to delay when a billable user is created.

Customer request.

Manually tested it.

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
+++ b/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
@@ -181,6 +181,15 @@ public class UrbanAirshipReactModule extends ReactContextBaseJavaModule {
     }
 
     /**
+     * If `channelCreationDelayEnabled` is enabled in the config, apps must call
+     * this method to enable channel creation.
+     */
+    @ReactMethod
+    public void enableChannelCreation() {
+        UAirship.shared().getPushManager().enableChannelCreation();
+    }
+
+    /**
      * Enables user notifications.
      *
      * @param promise The JS promise.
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -180,6 +180,15 @@ public void setUserNotificationsEnabled(boolean enabled) {
         UAirship.shared().getPushManager().setUserNotificationsEnabled(enabled);
     }
 
+    /**
+     * Enables channel creation if `channelCreationDelayEnabled` was
+     * enabled in the config.
```
BrianBatchelder: This is the method you want apps to call when they are ready to enable channel creation, right? So maybe the comment might be something like:
```
    * If `channelCreationDelayEnabled` is enabled in the config, apps must call 
    * this method to enable channel creation.
```

ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -52,6 +52,11 @@ - (RCTBridge *)bridge {
     [UAirship push].userPushNotificationsEnabled = enabled;
 }
 
+
```
BrianBatchelder: Is there a reason why these methods are missing comment headers?

rlepinski: None of them are defined

BrianBatchelder: I saw that. It's weird that they are defined in the Android code, but not the iOS code.

crow:
🍔

PR #204, opened 2019-07-30
[MOBILE-571] ReactNative message view

rlepinski:
### What do these changes do?
Adds the ability to embed a message view in a react native context.

Makes building a custom message center possible.

I ran it on both iOS and Android and verified both are able to load messages, handle close, etc...

Screenshot_1564437088

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
@@ -53,10 +53,11 @@ public class ReactMessageView extends UAWebView {
 
             if (error != null) {
                 notifyLoadError(message.getMessageId(), ERROR_MESSAGE_LOAD_FAILED, false);
-            } else {
-                message.markRead();
-                notifyLoadFinished(message.getMessageId());
+                return;
             }
+
+            message.markRead();
+            notifyLoadFinished(message.getMessageId());
         }
 
         @Override
--- a/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
@@ -1,0 +1,161 @@
+/* Copyright Airship and Contributors */
+
+package com.urbanairship.reactnative;
+
+import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.uimanager.events.RCTEventEmitter;
+import com.urbanairship.Cancelable;
+import com.urbanairship.UAirship;
+import com.urbanairship.richpush.RichPushInbox;
+import com.urbanairship.richpush.RichPushMessage;
+import com.urbanairship.widget.UAWebView;
+import com.urbanairship.widget.UAWebViewClient;
+
+public class ReactMessageView extends UAWebView {
+
+    static final String EVENT_LOAD_STARTED = "onLoadStarted";
+    static final String EVENT_LOAD_FINISHED = "onLoadFinished";
+    static final String EVENT_LOAD_ERROR = "onLoadError";
+    static final String EVENT_CLOSE = "onClose";
+
+    private static final String MESSAGE_ID_KEY = "message";
+    private static final String RETRYABLE_KEY = "retryable";
+    private static final String ERROR_KEY = "error";
+
+
+    private static final String ERROR_MESSAGE_NOT_AVAILABLE = "MESSAGE_NOT_AVAILABLE";
+    private static final String ERROR_FAILED_TO_FETCH_MESSAGE = "FAILED_TO_FETCH_MESSAGE";
+    private static final String ERROR_MESSAGE_LOAD_FAILED = "MESSAGE_LOAD_FAILED";
+
+    private RichPushMessage message;
+    private Cancelable fetchMessageRequest;
+
+    private WebViewClient webViewClient = new UAWebViewClient() {
+        private Integer error = null;
+
+        @Override
+        public void onPageFinished(WebView view, String url) {
+            super.onPageFinished(view, url);
+
+            if (message == null) {
+                return;
+            }
+
+            if (error != null) {
+                notifyError(message.getMessageId(), ERROR_MESSAGE_LOAD_FAILED, false);
+            } else {
+                message.markRead();
+                notifyLoadFinished(message.getMessageId());
+            }
+        }
+
+        @Override
+        public void onReceivedError(WebView view, int errorCode, String description, @Nullable String failingUrl) {
+            super.onReceivedError(view, errorCode, description, failingUrl);
+
+            if (message != null && failingUrl != null && failingUrl.equals(message.getMessageBodyUrl())) {
+                error = errorCode;
+            }
+        }
+
+        @Override
+        public void onClose(@NonNull WebView webView) {
+            if (message != null) {
+                notifyClose(message.getMessageId());
+            }
+        }
+    };
+
+    public ReactMessageView(@NonNull Context context) {
+        super(context);
+        setWebViewClient(webViewClient);
+    }
+
+    public void loadMessage(final String messageId) {
+        if (fetchMessageRequest != null) {
+            fetchMessageRequest.cancel();
+        }
+        this.message = null;
+
+        new Handler(Looper.getMainLooper()).post(new Runnable() {
+            @Override
+            public void run() {
+                startLoading(messageId);
+            }
+        });
+    }
+
+    void startLoading(final String messageId) {
+        notifyLoadStarted(messageId);
+
+        this.message = UAirship.shared().getInbox().getMessage(messageId);
+        if (this.message == null) {
+            fetchMessageRequest = UAirship.shared().getInbox().fetchMessages(new RichPushInbox.FetchMessagesCallback() {
+                @Override
+                public void onFinished(boolean success) {
+                    RichPushMessage message = UAirship.shared().getInbox().getMessage(messageId);
+                    if (!success) {
+                        notifyError(messageId, ERROR_FAILED_TO_FETCH_MESSAGE, true);
+                        return;
+                    } else if (message == null || message.isExpired()) {
+                        notifyError(messageId, ERROR_MESSAGE_NOT_AVAILABLE, false);
+                        return;
+                    }
+
+                    loadRichPushMessage(message);
+                }
+            });
+        } else {
+            if (this.message.isExpired()) {
+                notifyError(messageId, ERROR_MESSAGE_NOT_AVAILABLE, false);
+                return;
+            }
+            loadRichPushMessage(this.message);
+        }
+    }
+
+    private void notifyError(String messageId, String error, boolean retryable) {
+        WritableMap event = Arguments.createMap();
+        event.putString(MESSAGE_ID_KEY, messageId);
+        event.putBoolean(RETRYABLE_KEY, retryable);
+        event.putString(ERROR_KEY, error);
+        notify(EVENT_LOAD_ERROR, event);
+    }
+
+    private void notifyLoadFinished(String messageId) {
+        WritableMap event = Arguments.createMap();
+        event.putString(MESSAGE_ID_KEY, messageId);
+        notify(EVENT_LOAD_FINISHED, event);
+    }
+
+    private void notifyLoadStarted(String messageId) {
+        WritableMap event = Arguments.createMap();
+        event.putString(MESSAGE_ID_KEY, messageId);
+        notify(EVENT_LOAD_STARTED, event);
+    }
+
+    private void notifyClose(String messageId) {
+        WritableMap event = Arguments.createMap();
+        event.putString(MESSAGE_ID_KEY, messageId);
+        notify(EVENT_CLOSE, event);
+    }
+
+    private void notify(String eventName, WritableMap event) {
+        ReactContext reactContext = (ReactContext) getContext();
+        reactContext.getJSModule(RCTEventEmitter.class).receiveEvent(
+                getId(),
+                eventName,
+                event);
+    }
+}
--- a/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
@@ -33,7 +33,6 @@ public class ReactMessageView extends UAWebView {
     private static final String RETRYABLE_KEY = "retryable";
     private static final String ERROR_KEY = "error";
 
-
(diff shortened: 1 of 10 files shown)


android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -0,0 +1,161 @@
...
+    static final String EVENT_LOAD_STARTED = "onLoadStarted";
+    static final String EVENT_LOAD_FINISHED = "onLoadFinished";
+    static final String EVENT_LOAD_ERROR = "onLoadError";
+    static final String EVENT_CLOSE = "onClose";
+
+    private static final String MESSAGE_ID_KEY = "message";
+    private static final String RETRYABLE_KEY = "retryable";
+    private static final String ERROR_KEY = "error";
+
+
```
BrianBatchelder: whitespace

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -0,0 +1,161 @@
...
+                public void onFinished(boolean success) {
+                    RichPushMessage message = UAirship.shared().getInbox().getMessage(messageId);
+                    if (!success) {
+                        notifyError(messageId, ERROR_FAILED_TO_FETCH_MESSAGE, true);
+                        return;
+                    } else if (message == null || message.isExpired()) {
+                        notifyError(messageId, ERROR_MESSAGE_NOT_AVAILABLE, false);
+                        return;
+                    }
+
```
BrianBatchelder: whitespace

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
```diff
@@ -0,0 +1,59 @@
...
+    protected ReactMessageView createViewInstance(@Nonnull ThemedReactContext reactContext) {
+        return new ReactMessageView(reactContext);
+    }
+
+    @ReactProp(name = "messageId")
+    public void setMessageId(ReactMessageView view, @Nullable String messageId) {
+        view.loadMessage(messageId);
+    }
+
+
```
BrianBatchelder: whitespace

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,178 @@
...
+        self.webView.allowsLinkPreview = ![UAirship messageCenter].disableMessageLinkPreviewAndCallouts;
+        self.webView.configuration.dataDetectorTypes = WKDataDetectorTypeAll;
+        [self addSubview:self.webView];
+    }
+    return self;
+}
+
+- (void)layoutSubviews {
+  [super layoutSubviews];
+  self.webView.frame = self.bounds;
```
BrianBatchelder: indentation

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -0,0 +1,161 @@
...
+        } else {
+            if (this.message.isExpired()) {
+                notifyError(messageId, ERROR_MESSAGE_NOT_AVAILABLE, false);
+                return;
+            }
+            loadRichPushMessage(this.message);
+        }
+    }
+
+    private void notifyError(String messageId, String error, boolean retryable) {
```
BrianBatchelder: On iOS, it looks like this method is called `onLoadError`. Maybe `notifyLoadError()` here?

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,178 @@
...
+    [[UAirship inboxUser] getUserData:^(UAUserData *userData) {
+        UA_STRONGIFY(self)
+        NSString *auth = [UAUtils userAuthHeaderString:userData];
+        [request setValue:auth forHTTPHeaderField:@"Authorization"];
+        [self.webView loadRequest:request];
+    } queue:dispatch_get_main_queue()];
+}
+
+- (void)webView:(WKWebView *)wv decidePolicyForNavigationResponse:(WKNavigationResponse *)navigationResponse decisionHandler:(void (^)(WKNavigationResponsePolicy))decisionHandler {
+
```
BrianBatchelder: whitespace

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
```diff
@@ -0,0 +1,59 @@
...
+    public Map getExportedCustomBubblingEventTypeConstants() {
+
```
BrianBatchelder: whitespace

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,178 @@
...
+    decisionHandler(WKNavigationResponsePolicyAllow);
+
+}
+
+- (void)webView:(WKWebView *)wv didFinishNavigation:(WKNavigation *)navigation {
+    if ([UAirship messageCenter].disableMessageLinkPreviewAndCallouts) {
+        [self.webView evaluateJavaScript:@"document.body.style.webkitTouchCallout='none';" completionHandler:nil];
+    }
+
+    // Mark message as read after it has finished loading
```
BrianBatchelder: This comment implies to me that `markMessageReadWithCompletionHandler` won't be called until after the message has finished loading, which is a bit confusing given the next `if`. Perhaps just "Mark unread messages as read"?

ios/UARCTModule/UARCTMessageViewManager.m
```diff
@@ -0,0 +1,23 @@
+/* Copyright Airship and Contributors */
+
+#import "UARCTMessageViewManager.h"
+#import "UARCTMessageView.h"
+
+#import <React/RCTBridge.h>
+#import <React/RCTUIManager.h>
+
```
BrianBatchelder: whitespace

js/UAMessageView.js
```diff
@@ -0,0 +1,82 @@
...
+   */
+  onLoadError:  PropTypes.func,
+
+  /**
+   * The ID of a message to load.
+   */
+  messageId: PropTypes.string
+};
+
+
```
BrianBatchelder: whitespace

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.Nonnull;
+
+public class ReactMessageViewManager extends SimpleViewManager<ReactMessageView> {
+
+    public static final String REACT_CLASS = "UARCTMessageView";
```
marc-scig: Is this correct?

ios/UARCTModule/UARCTMessageViewManager.m
```diff
@@ -0,0 +1,23 @@
...
+
```
BrianBatchelder: whitespace

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
marc-scig: I only ask because I notice it's a class name in the iOS native code

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
```diff
@@ -0,0 +1,57 @@
...
+    protected ReactMessageView createViewInstance(@Nonnull ThemedReactContext reactContext) {
+        return new ReactMessageView(reactContext);
+    }
+
+    @ReactProp(name = "messageId")
+    public void setMessageId(ReactMessageView view, @Nullable String messageId) {
+        view.loadMessage(messageId);
+    }
+
+    public Map getExportedCustomBubblingEventTypeConstants() {
```
marc-scig: what is this section for? it looks like it has a fair amount of magic strings

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -0,0 +1,159 @@
...
+    static final String EVENT_LOAD_STARTED = "onLoadStarted";
+    static final String EVENT_LOAD_FINISHED = "onLoadFinished";
+    static final String EVENT_LOAD_ERROR = "onLoadError";
+    static final String EVENT_CLOSE = "onClose";
+
+    private static final String MESSAGE_ID_KEY = "message";
+    private static final String RETRYABLE_KEY = "retryable";
+    private static final String ERROR_KEY = "error";
+
+    private static final String ERROR_MESSAGE_NOT_AVAILABLE = "MESSAGE_NOT_AVAILABLE";
```
marc-scig: Are these values intended to be keys used for constructing human readable errors, or error messages themselves?

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -0,0 +1,159 @@
...
+
+        @Override
+        public void onPageFinished(WebView view, String url) {
+            super.onPageFinished(view, url);
+
+            if (message == null) {
+                return;
+            }
+
+            if (error != null) {
```
crow: probably could be cleaner/more consistent with an early return

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,175 @@
...
+
+    self.fetchMessagesDisposable = [[UAirship inbox].messageList retrieveMessageListWithSuccessBlock:^{
+        dispatch_async(dispatch_get_main_queue(), ^{
+            UA_STRONGIFY(self)
+
+            UAInboxMessage *message = [[UAirship inbox].messageList messageForID:messageID];
+            if (message && !message.isExpired) {
+                [self requestMessageBody:message];
+            } else {
+                if (self.onLoadError) {
```
crow: bring this block out into an else if?

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,175 @@
...
+@implementation UARCTMessageView
+
+- (instancetype) init {
+    self = [super initWithFrame:CGRectZero];
+    if (self) {
+        self.nativeBridge = [[UAWKWebViewNativeBridge alloc] init];
+        self.nativeBridge.forwardDelegate = self;
+        self.webView = [[WKWebView alloc] initWithFrame:self.bounds];
+        self.webView.navigationDelegate = self.nativeBridge;
+        self.webView.allowsLinkPreview = ![UAirship messageCenter].disableMessageLinkPreviewAndCallouts;
```
marc-scig: does this mean that allows link preview defaults to YES since in the message center it defaults to NO? Is the switch something someone can or should flip from the js side?

ios/UARCTModule/UARCTMessageView.m
```diff
@@ -0,0 +1,175 @@
...
+    }
+    return self;
+}
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    self.webView.frame = self.bounds;
+}
+
+- (void)setMessageID:(NSString *)messageID {
```
marc-scig: where does this get called from?

marc-scig: also, is there a need for the property? could you just pass it to loadMessage and mark it non-null?

ios/UARCTModule/UARCTMessageViewManager.m
```diff
@@ -0,0 +1,23 @@
...
+@implementation UARCTMessageViewManager
+
+RCT_EXPORT_VIEW_PROPERTY(onLoadStarted, RCTDirectEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onLoadFinished, RCTDirectEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onLoadError, RCTDirectEventBlock)
+RCT_EXPORT_VIEW_PROPERTY(onClose, RCTDirectEventBlock)
+RCT_REMAP_VIEW_PROPERTY(messageId, messageID, NSString)
+RCT_EXPORT_MODULE(UARCTMessageView)
+
+- (UIView *)view {
```
marc-scig: I remember this being a problem when I looked into this eons ago. How does this work? Like where does the view live once you've created it, and what view controller does it attach to?

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
rlepinski: This is just the glue from the JS side to the view manager. I reused the iOS class name so we could have a single .js file

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
rlepinski: Really confused by this too. Before adding this code it was throwing an exception about an unexpected event. It seems react needs some sort of registry on Android - https://facebook.github.io/react-native/docs/native-components-android#events

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
rlepinski: Sorta both. When an error happens we send an event:

```
"error": "MESSAGE_NOT_AVAILABLE",
"message": "bloop",
"retryable": false
```

ios/UARCTModule/UARCTMessageView.m
rlepinski: This is code from our built in viewcontroller. We could expose it if someone wants it.

ios/UARCTModule/UARCTMessageView.m
rlepinski: The message ID is a property on the js view object

```
<UAMessage
   messageId={state.messageId}
/>
```

To make that work, i had to use the macro to map the view property to the view:
```
RCT_REMAP_VIEW_PROPERTY(messageId, messageID, NSString)
```

I don't know if that accepts a method instead of a property, but since its a property in react it might as well be a property in obj-c

ios/UARCTModule/UARCTMessageViewManager.m
rlepinski: The docs did not go into much detail here. The ViewManager is singleton and this is called for every `UAMessageView` instance in the react view layout. The view will have its frame  overriden, so you can construct it with CGRectZero.

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
rlepinski: It connects this line `var UARCTMessageView = requireNativeComponent('UARCTMessageView', UAMessageView);
`

android/src/main/java/com/urbanairship/reactnative/ReactMessageViewManager.java
marc-scig: weird

android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
marc-scig: ok

ios/UARCTModule/UARCTMessageView.m
marc-scig: I guess then what I really mean is should it just have a hardcoded YES value here instead of delegating to something that will always say YES

PR #205, opened 2019-07-30
[MOBILE-809] Skip checking for message when displaying message with ID

rlepinski:
### What do these changes do?
Skips the message look up with a message is being displayed from an ID. The message views know how to handle refreshing the inbox if the message is not available.

Makes the API easier to use.

Manually tested both iOS and Android.


ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -372,27 +372,19 @@ - (RCTBridge *)bridge {
 
     UAInboxMessage *message = [[UAirship inbox].messageList messageForID:messageId];
```
BrianBatchelder: Looks like you don't need the message anymore.

PR #208, opened 2019-08-01
Release 3.2.0

rlepinski:
Release 3.2.0. Will follow up with a 4.0.0 release for AndroidX and ReactNative 0.6

Only take a look at the changelog, that really is the only new thing. Everything else is from the 4.0 branch. CI is going to fail because its setup for the 4.0 branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+Version 3.2.0 - August 1, 2019
+==============================
+- Fixed `isDeleted` always true for iOS inbox messages.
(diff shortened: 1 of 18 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+Version 3.2.0 - August 1, 2019
+==============================
+- Fixed `isDeleted` always true for iOS inbox messages.
```
crow: always being set to? Doesn't read quite right

PR #213, opened 2019-08-23
[MOBILE-832] use documentation.js

marc-scig:
JSDoc has been causing us some pain, due to conflicts with our required babel plugin version. In the meanwhile, we've decided to try using [documentation.js](https://documentation.js.org), which supports JSDoc syntax, and supports flow out the box without need for babel. It's not quite as fancy as JSDoc, so some stuff needed a few additional annotations to get it to show up correctly.

Note: until we fix our mobile-docs jenkins situation, we won't be able to publish these in our bitrise job. I'll attach some screenshots so everyone can get a feel for what the output looks like.

Screen Shot 2019-08-23 at 2 31 22 PM
Screen Shot 2019-08-23 at 2 31 33 PM
Screen Shot 2019-08-23 at 2 31 44 PM


crow:
🍔

Issue #212, opened 2019-08-22
UAMessageView covers top of screen for a few seconds


pflenker:
# Preliminary Info

### What Airship dependencies are you using?
3.2.1

# Report

### What unexpected behavior are you seeing?
When using UAMessageView, any content above the view (Navigation bar, Close button...) is covered by a white view for a few seconds/milliseconds before it appears.

### What is the expected behavior?
Everything outside of UAMessageView stays unchanged.

### What are the steps to reproduce the unexpected behavior?
Simplest test case: Create a structure like this:
```jsx
<View>
  <Text>Foo</Text>
  <UAMessageView messageId={42} />
</View>
```
The text will be covered for a bit even after the message has loaded before appearing again. In our case, a whole navigation bar is affected, including the button to dismiss the view.

pflenker:
I am on React Native 0.59.10  and Urban Airship 3.2.1.

Here is minimal code to reproduce the issue:

```jsx
import React, { useState, useEffect } from 'react';
import {
  FlatList, TouchableOpacity, Text, View,
} from 'react-native';

import { UrbanAirship, UAMessageView } from 'urbanairship-react-native';


const UAInboxScreen = () => {
  const [items, setItems] = useState([]);
  const [messageId, setMessageId] = useState(undefined);
  async function loadItems() {
    const loadedItems = await UrbanAirship.getInboxMessages();
    setItems(loadedItems);
  }
  useEffect(() => {
    loadItems();
  }, []);
  if (messageId) {
    return (
      <View>
        <Text style={{
          color: 'red',
          fontSize: 20,
        }}
        >
          Foo
        </Text>
        <UAMessageView
          style={{
            height: 200,
          }}
          messageId={messageId}
        />
      </View>
    );
  }
  return (
    <View>
      <FlatList
        data={items}
        renderItem={({ item }) => (
          <TouchableOpacity onPress={() => {
            setMessageId(item.id);
          }}
          >
            <Text>
              {item.id}
            </Text>
          </TouchableOpacity>
        )}
        keyExtractor={item => item.id}
      />
    </View>
  );
};


export default UAInboxScreen;
```

Here is a [video](https://github.com/urbanairship/react-native-module/files/3550603/recordedVideo.zip)

pflenker:
on iOS, the UAMessageView won't even display, with the following error : 

[...]

PR #214, opened 2019-09-03
Release 3.2.2

rlepinski:
Fixes GH-212. The main issue seemed to be embedding a webview directly in a react view stack. IU was able to work around the issue by wrapping the UAWebView in a framelayout. Also added missing files to the xcode project on iOS.


rlepinski:
CI is going to fail on the older branch

PR #223, opened 2019-10-04
MOBILE-807 react native testing

marc-scig:
The test framework I'm using here, jest, is developed by Facebook and is considered the de facto choice for react development. It's a little different from what we're used to, particularly where mocking is concerned (and especially when native modules are being used). I worked out some tests for UAEventEmitter, TagGroupEditor and UACustomEvent as an example of how to go about this for future enhancements. Tests can be run with a script by invoking `npm test`.

The main UrbanAirshipModule still needs tests but should be relatively straightforward with the existing ones as a guide. We should also be able to do some form of snapshot testing for the message view component, but it's not yet clear to me how much that will buy us, and it's an esoteric enough technique that we may as well save it for a follow up ticket.

Edit: should be mergeable now. I fixed the event emitter test by fully mocking out the relevant bits instead of relying on concrete implementations, which the newest version of react broke for reasons I don't fully understand. It's better this way. Also added the `__tests__` directory to .npmignore at ryan's suggestion.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 example
 .github
 scripts
+js/__tests__
--- a/js/__tests__/TagGroupEditor.test.js
+++ b/js/__tests__/TagGroupEditor.test.js
@@ -1,0 +1,45 @@
+/* Copyright Airship and Contributors */
+
+const TagGroupEditor = require("../TagGroupEditor");
+
+describe("TagGroupEditor Tests", () => {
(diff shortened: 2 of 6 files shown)


BrianBatchelder:
Looks good. I think it is good to prioritize using the de facto standard. The format is not familiar, but we are used to learning new things. I do find it pretty readable, which is good.

PR #226, opened 2019-10-25
[MOBILE-958] Added the ability to turn on Autobadging in React Native

Jahdeh:
https://urbanairship.atlassian.net/browse/MOBILE-958

### What do these changes do?
These changes add the ability to enable Autobadging in the React Native SDK.

Folks need a way to turn on Autobadging in this SDK just like the iOS SDK.

I called the methods from App.js in the example and made sure I could return the same boolean from isAutobadgeEnabled that I set in setAutobadgeEnabled. 

To reproduce:
1. Pull this branch and switch to it
2. Somewhere in App.js (I used the constructor), call setAutobadgeEnabled and pass in true.
3. Call isAutobadgeEnbled and add .then to the end, passing in a lambda function with an argument.
```
UrbanAirship.setAutobadgeEnabled(true);
UrbanAirship.isAutobadgeEnabled().then(result => console.log(result));
```
4. Print out the result and confirm that true is the boolean value returned.

### DIFF
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -495,7 +495,7 @@ class UrbanAirship {
   }
 
   /**
-   * Sets Autobadging on iOS, either to true or false. Badging is not supported for Android.
+   * Enables or disables autobadging on iOS. Badging is not supported for Android.
    * 
    * @param {boolean} enabled Whether or not to enable Autobadging.
    */
@@ -516,7 +516,8 @@ class UrbanAirship {
     if (Platform.OS == 'ios') {
       return UrbanAirshipModule.isAutobadgeEnabled();
     } else {
-      console.log("This feature is not supported on this platform.")
+      console.log("This feature is not supported on this platform.");
+      return new Promise(resolve => resolve(false));
     }
   }
 
--- a/js/UrbanAirship.js
+++ b/js/UrbanAirship.js
@@ -495,10 +495,10 @@ class UrbanAirship {
   }
 
   /**
-  * Sets Autobadging on iOS, either to true or false. Badging is not supported for Android.
-  * 
-  * @param {boolean} enabled Whether or not to enable Autobadging.
-  */
+   * Sets Autobadging on iOS, either to true or false. Badging is not supported for Android.
+   * 
+   * @param {boolean} enabled Whether or not to enable Autobadging.
+   */
   static setAutobadgeEnabled(enabled: boolean) {
     if (Platform.OS == 'ios') {
       UrbanAirshipModule.setAutobadgeEnabled(enabled);
@@ -513,10 +513,11 @@ class UrbanAirship {
    * @return {Promise.<boolean>} A promise with the result, either true or false.
    */
   static isAutobadgeEnabled(): Promise<boolean> {
-    if (Platform.OS != 'ios') {
+    if (Platform.OS == 'ios') {
+      return UrbanAirshipModule.isAutobadgeEnabled();
+    } else {
       console.log("This feature is not supported on this platform.")
     }
-    return UrbanAirshipModule.isAutobadgeEnabled();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


js/UrbanAirship.js
```diff
@@ -494,6 +494,31 @@ class UrbanAirship {
     return UrbanAirshipModule.isQuietTimeEnabled();
   }
 
+  /**
+  * Sets Autobadging on iOS, either to true or false. Badging is not supported for Android.
+  * 
+  * @param {boolean} enabled Whether or not to enable Autobadging.
+  */
```
BrianBatchelder: Need an extra space at the beginning of the line.

js/UrbanAirship.js
```diff
@@ -494,6 +494,31 @@ class UrbanAirship {
...
+    }
+}
+
+  /**
+   * Checks to see if Autobadging on iOS is enabled. Badging is not supported for Android.
+   * 
+   * @return {Promise.<boolean>} A promise with the result, either true or false.
+   */
+  static isAutobadgeEnabled(): Promise<boolean> {
+    if (Platform.OS != 'ios') {
```
BrianBatchelder: Could you flip this if/else to match the one in `setAutobadgeEnabled()`?

js/UrbanAirship.js
Jahdeh: Okie doke! So add a space after the */ on 501?

js/UrbanAirship.js
Jahdeh: Can do! Word of warning, I was matching the pattern that was set by setBadgeNumber and getBadgeNumber. Did you want me to change those as well in this PR?

js/UrbanAirship.js
BrianBatchelder: Before. The asterisks should all line up, like in the next method.

js/UrbanAirship.js
BrianBatchelder: Nah, just do this one. It's mostly just readability.

js/UrbanAirship.js
Jahdeh: Ah okay! I see it now! Thank you!

js/UrbanAirship.js
```diff
@@ -494,6 +494,32 @@ class UrbanAirship {
     return UrbanAirshipModule.isQuietTimeEnabled();
   }
 
+  /**
+   * Sets Autobadging on iOS, either to true or false. Badging is not supported for Android.
```
marc-scig: I'd recommend rephrasing the first bit as "Enables or disables autobadging on iOS."

js/UrbanAirship.js
```diff
@@ -494,6 +494,32 @@ class UrbanAirship {
...
+  /**
+   * Checks to see if Autobadging on iOS is enabled. Badging is not supported for Android.
+   * 
+   * @return {Promise.<boolean>} A promise with the result, either true or false.
+   */
+  static isAutobadgeEnabled(): Promise<boolean> {
+    if (Platform.OS == 'ios') {
+      return UrbanAirshipModule.isAutobadgeEnabled();
+    } else {
+      console.log("This feature is not supported on this platform.")
```
marc-scig: what happens if someone calls this on android, will it return undefined? if so do we want it to return false?

Jahdeh: I can certainly add that, or maybe return null since there's nothing to really do if it's android?

js/UrbanAirship.js
Jahdeh: Okie doke!

js/UrbanAirship.js
marc-scig: I'd suggest returning false. since the method is still technically callable from the JS side on android, that will make its return signature consistent, and "unavailable" just means "permanently disabled"

Jahdeh: Can do!

PR #228, opened 2019-11-06
[MOBILE-1006] Fixing MessageView callbacks

Apekka:
### What do these changes do?
Theses changes fix a bug in the ReactMessageView class, where the variable "message" was re-created.

To fix an issue.

I sent a message to my device, and checked if everything is displaying correctly and that there is no crash.

screen

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
@@ -106,7 +106,7 @@ public class ReactMessageView extends FrameLayout implements LifecycleEventListe
             fetchMessageRequest = UAirship.shared().getInbox().fetchMessages(new RichPushInbox.FetchMessagesCallback() {
                 @Override
                 public void onFinished(boolean success) {
-                    RichPushMessage message = UAirship.shared().getInbox().getMessage(messageId);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/urbanairship/reactnative/ReactMessageView.java
```diff
@@ -106,7 +106,7 @@ void startLoading(final String messageId) {
             fetchMessageRequest = UAirship.shared().getInbox().fetchMessages(new RichPushInbox.FetchMessagesCallback() {
                 @Override
                 public void onFinished(boolean success) {
-                    RichPushMessage message = UAirship.shared().getInbox().getMessage(messageId);
+                    message = UAirship.shared().getInbox().getMessage(messageId);
```
BrianBatchelder: Do we need this mix of `this.message` and `message`? It's pretty confusing, and may have contributed to the problem. FWIW: I don't really care which one we use, but, if forced to choose, I'd say use `this.message`.

rlepinski: `this.message` will not work in this context, that would refer to the callback. You would have to use `ReactMessageView.this.message`.

PR #234, opened 2019-12-14
[MOBILE-1131] add module version tracking to ios and android

crow:
Adds module tracking to ios and android and does some cleanup for the release


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2010-2019 Airship and Contributors
+Copyright 2010-2020 Airship and Contributors
```
rlepinski: living in the future

PR #238, opened 2020-01-02
Release 6.0.1

crow:
Npm packages installed from a local source (i.e.`"file:../package/"` in the package.json) don't strip out files included in the .npmignore. This made my tests of the last plugin update unreliable. 

### What do these changes do?
Includes the script/sync_version.sh script in the published npm package

Module install fails without the `sync_version.sh`script.

Ran `npm pack` and then checked the contents of the generated tarball to ensure the required script was in place.

tarball_check

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,4 @@
 example
 .github
-scripts
 js/__tests__
+scripts/run_ci_tasks.sh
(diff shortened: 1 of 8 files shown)


oristanovic:
👍

PR #248, opened 2020-02-12
[MOBILE-1270] Add number attribute support to React Native

Apekka:
### What do these changes do?
Adding number support to the React Native "setAttribute" method

To send number values to attributes

By installing the sample app, adding buttons that call the method "setAttribute" with a string value, and an other one calling it with a numeric (double) value. And see that they are correctly sent to the server.


js/AttributeEditor.js
```diff
@@ -21,7 +21,7 @@ class AttributeEditor {
    * @param {string} name The attribute name.
    * @return {AttributeEditor} The attribute editor instance.
    */
-  setAttribute(value: string, name: string): AttributeEditor {
+  setAttribute(name: string, value: string|number): AttributeEditor {
```
crow: blessu

PR #252, opened 2020-02-21
[MOBILE-1335] Simplify release workflow

rlepinski:
### What do these changes do?

Uploads docs to GCS instead of mobile, adds a version check on the tag, and simplifies the release flow to be two parts (build/ci vs publish)

We want to remove the use of the mobile-docs repo

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,30 +6,33 @@ on:
       - "[0-9]+.[0-9]+.[0-9]+*"
 
 jobs:
-
-  android:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -41,49 +44,28 @@ jobs:
...
-    needs: [android, ios, publish_docs, publish]
-    steps:
       - name: Slack Notification
         uses: homoluctus/slatify@master
-        if: always()
+        if: success()
         with:
           type: ${{ job.status }}
-          job_name: ":raised_hands: Airship React native Plugin Released! :raised_hands:"
+          job_name: "Airship React native Plugin Released!"
```
marc-scig: aw, no emoji?

rlepinski: dont put that pressure on me

PR #253, opened 2020-02-21
MOBILE-1269 Workspace/Yarn Refactor

marc-scig:
This PR restructures the repo to resemble https://github.com/react-native-community/react-native-slider, which has been promoted as a standard example of how to properly configure a react-native monorepo containing a module and example app. One of the reasons for doing this now is that our existing structure causes problems with version 3.x of the cli tools. Another is to make it more legible to the wider react-native community.

The biggest change is moving from npm to yarn. Yarn uses the same package.json format, with a few minor additions to facilitate workspaces, which will let us publish modules locally for use in a sample app without setting our hair on fire. This also means moving some files around, so that the module source lives in a src/ directory off the root. One advantage of the workspace approach is that both the module and example can have their own package.json, and therefore their own dependencies. So several dependencies that were once tied to the module (gesture, tab bar, etc) are now strictly part of the example app.

To make it possible to edit module source code without having to reinstall every time you run the example, this solution uses yarn link, which is similar to npm link except that it appears to play nicely with Metro. I've updated the dev readme to reflect those changes.

The yarn command alone is sufficient to run scripts, and either `yarn react-native` or `yarn scriptname`, where scriptname references react-native, will invoke the react native cli contained in node_modules. Since we're on react-native 0.61 now, this means 3.x cli tools, and globally installed cli tools are effectively obsolete for our purposes.

This should be properly rebased now, and I updated the CI scripts accordingly. CI tasks are running fine locally, and I think the deploy workflow should be fine though we'll find out for sure soon enough.


DEV_README.md
```diff
@@ -14,10 +14,24 @@ brew install watchman
...
 ```
-npm install
+yarn install
 ```
 
+3) Link urbanairship-react-native
+
+Execute the following command in src/
+
+yarn link
```
BrianBatchelder: Put in code block.

DEV_README.md
```diff
@@ -14,10 +14,24 @@ brew install watchman
...
+
+Next, execute the following command in example/
+
+yarn link urbanairship-react-native
```
BrianBatchelder: Put in code block.

example/android/app/build.gradle
```diff
@@ -76,7 +76,9 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "index.js"
+    entryFile: "example/index.js",
+    // TODO: remove once @esemesek's patch is merged
```
BrianBatchelder: Maybe more details on identifying the patch, in case you're not the one to fix this.

marc-scig: I copied this directly from the slider example, and I'll be honest I don't fully understand why it's necessary

DEV_README.md
marc-scig: done

DEV_README.md
marc-scig: done

example/android/app/build.gradle
marc-scig: turns out it wasn't necessary

marc-scig:
@BrianBatchelder yarn uses package.json, but has its own lockfile (yarn.lock). so the package-lock.json file is no more.

PR #271, opened 2020-05-12
[MOBILE-1567] Named user attributes

oristanovic:
### What do these changes do?
Add names user attributes support.

Ti allow customers using React Native to use this feature.

Manual tests in sample apps.

### DIFF
--- a/src/js/UrbanAirship.js
+++ b/src/js/UrbanAirship.js
@@ -327,25 +327,13 @@ class UrbanAirship {
   /**
    * Creates an editor to modify the named user attributes.
    *
-   * @return {Promise.<AttributeEditor, Error>}  A promise that returns an attribute editor instance if resolved, or an Error if the
-   * named user doesn't exist.
+   * @return {AttributeEditor} A attribute editor instance.
    */
-  static editNamedUserAttributes(): Promise {
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


src/js/UrbanAirship.js
```diff
@@ -324,6 +324,29 @@ class UrbanAirship {
     });
   }
 
+  /**
+   * Creates an editor to modify the named user attributes.
+   *
+   * @return {Promise.<AttributeEditor, Error>}  A promise that returns an attribute editor instance if resolved, or an Error if the
```
BrianBatchelder, line 330: The SDKs allow attributes to be set before setting a named user id. The attributes are uploaded once a named user id has been set. Why did you decide to force the named user id to be set before setting attributes?

oristanovic: I was following the ticket asking to return an error in the callback if the named user is not set.

BrianBatchelder: Well, that's weird. Why would we do that in frameworks if the SDKs don't do it? I feel like we should get @rlepinski to clarify.

oristanovic:
> I think we might have to use the `--alow-warnings` flag in the pod lint step to work around the MobileCoreServices name change

It's because the SDK is not updated yet, I pushed the podspec change to make the checks pass.

PR #273, opened 2020-05-14
[MOBILE-1585-ts] update typescript definitions, fix jsdoc

marc-scig:
This ticket was originally to add typescript definitions for the data collection methods, but I did an audit and found we were missing several other things. I also cleaned up some problematic parts of the documentation in both TS and JS. We will want to take extra care to make sure that the new additions in other recent tickets are updated in in index.d.ts.

Tested manually with a bespoke typescript app.

### DIFF
--- a/src/js/AttributeEditor.js
+++ b/src/js/AttributeEditor.js
@@ -14,10 +14,10 @@ class AttributeEditor {
   }
 
   /**
-   * Adds string attribute.
+   * Adds an attribute.
    * @instance
    * @memberof AttributeEditor
-   * @function setString
+   * @function setAttribute
    *
    * @param {string} value The attribute value.
    * @param {string|number|Date} name The attribute name.
@@ -45,7 +45,7 @@ class AttributeEditor {
   }
 
   /**
-   * Removes the attribute.
+   * Removes an attribute.
    * @instance
    * @memberof AttributeEditor
    * @function removeAttribute
--- a/src/js/UrbanAirship.js
+++ b/src/js/UrbanAirship.js
@@ -175,14 +175,14 @@ class UrbanAirship {
 
   /**
    * Global data collection flag. Enabled by default, unless `dataCollectionOptInEnabled`
-   * is set to `YES` in AirshipConfig.
+   * is set to `YES` in AirshipConfig.plist on iOS, and `true` in airshipconfig.properties on Android.
    * When disabled, the device will stop collecting and sending data for named user, events,
    * tags, attributes, associated identifiers, and location from the device.
    *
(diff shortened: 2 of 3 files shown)


src/js/index.d.ts
```diff
@@ -27,15 +27,21 @@ declare class UACustomEvent {
   ): void;
 }
 
-declare interface Operation {
+declare interface TagGroupOperation {
```
marc-scig, line 30: This could technically be interpreted as an interface change, but it's just a typescript definition for object parameters, and it's also for the module, not for public use. I don't think it's a problem.

src/js/AttributeEditor.js
```diff
@@ -14,13 +14,13 @@ class AttributeEditor {
   }
 
   /**
-   * Adds string attribute.
+   * Adds an attribute.
    * @instance
    * @memberof AttributeEditor
    * @function setString
```
BrianBatchelder, line 20:
```suggestion
   * @function setAttribute
```

src/js/UrbanAirship.js
```diff
@@ -175,14 +175,13 @@ class UrbanAirship {
...
   /**
    * Global data collection flag. Enabled by default, unless `dataCollectionOptInEnabled`
-   * is set to `YES` in AirshipConfig.
+   * is set to `YES` in AirshipConfig.plist on iOS, and `true` in airshipconfig.properties on Android.
    * When disabled, the device will stop collecting and sending data for named user, events,
    * tags, attributes, associated identifiers, and location from the device.
    *
-   * Push notifications will continue to work only if `UrbanAirshipModule.setPushTokenRegistrationEnabled`
-   * has been explicitly set to `YES`, otherwise it will default to the current state  of `isDataCollectionEnabled`.
+   * Push notifications will continue to work only if `UrbanAirship.setPushTokenRegistrationEnabled`
```
BrianBatchelder, line 182: I think you lost a line here:
```suggestion
   * Push notifications will continue to work only if `UrbanAirship.setPushTokenRegistrationEnabled`
   * has been explicitly set to `YES`, otherwise it will default to the current state  of `isDataCollectionEnabled`.
```

src/js/index.d.ts
```diff
@@ -446,6 +538,20 @@ export class UrbanAirship {
    */
   static isQuietTimeEnabled(): Promise<boolean>;
 
+  /**
+   * Sets the badge number for iOS. Badging is not supported for Android.
+   *
+   * @param {number} badgeNumber specified badge to set.
```
BrianBatchelder, line 544: Did you lose `setBadgeNumber`?

PR #274, opened 2020-05-14
[MOBILE-1564] Remove property restrictions for custom events

BrianBatchelder:
### What do these changes do?
Allow custom events to accept any JSON.

Events need JSON.

Unit tests. Local testing using modified sample app.

I was pleasantly surprised that everything _just worked_ once I extend the method signatures.

### DIFF
--- a/src/js/UACustomEvent.js
+++ b/src/js/UACustomEvent.js
@@ -39,9 +39,9 @@ class UACustomEvent {
      * Adds a property to the custom event.
      *
      * @param {string} name The property name.
-     * @param {string|number|boolean|string[]} value The property value.
+     * @param {string|number|boolean|string[]|Object} value The property value.
      */
-    addProperty(name: string, value: string | number | boolean | Array<string>) {
+    addProperty(name: string, value: string | number | boolean | Array<string> | Object) {
       this._properties[name] = value;
     }
 }
--- a/src/js/UACustomEvent.js
+++ b/src/js/UACustomEvent.js
@@ -39,7 +39,7 @@ class UACustomEvent {
      * Adds a property to the custom event.
      *
      * @param {string} name The property name.
-     * @param {string|number|boolean|string[]|Object} value The property value.
+     * @param {string|number|boolean|string[]|Record<string, any>} value The property value.
      */
     addProperty(name: string, value: string | number | boolean | Array<string> | Record<string, any>) {
       this._properties[name] = value;
--- a/src/js/UACustomEvent.js
+++ b/src/js/UACustomEvent.js
@@ -3,6 +3,10 @@
 // @flow
 'use strict';
 
+type JsonValue = string | number | boolean | null | JsonMap | JsonArray;
+interface JsonMap {  [key: string]: JsonValue; }
+interface JsonArray extends Array<JsonValue> {}
+
 /**
  * Urban Airship Custom events
  **/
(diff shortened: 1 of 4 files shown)


src/js/index.d.ts
```diff
@@ -19,11 +19,11 @@ declare class UACustomEvent {
    * Adds a property to the custom event.
    *
    * @param {string} name The property name.
-   * @param {string|number|boolean|string[]} value The property value.
+   * @param {string|number|boolean|string[]|Object} value The property value.
```
marc-scig, line 22: in typescript we should be using the non-primitive `object` type instead of `Object`: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types

marc-scig: I also think we can probably afford to be more strict with the typesystem here and constrain the keys and values the object can contain.

marc-scig: e.g.

```
type JsonValue = string | number | boolean | null | JsonMap | JsonArray;
interface JsonMap {  [key: string]: JsonValue; }
interface JsonArray extends Array<JsonValue> {}
```

src/js/UACustomEvent.js
```diff
@@ -39,9 +39,9 @@ class UACustomEvent {
      * Adds a property to the custom event.
      *
      * @param {string} name The property name.
-     * @param {string|number|boolean|string[]} value The property value.
+     * @param {string|number|boolean|string[]|Object} value The property value.
```
marc-scig, line 42: we're missing array in the param field

src/js/UACustomEvent.js
```diff
@@ -39,9 +39,9 @@ class UACustomEvent {
...
      */
-    addProperty(name: string, value: string | number | boolean | Array<string>) {
+    addProperty(name: string, value: string | number | boolean | Array<string> | Object) {
```
marc-scig, line 44: is it still true that only string arrays are allowed?

BrianBatchelder: I don't know. But JsonValue should cover any array we would accept, right?

src/js/index.d.ts
BrianBatchelder: Done.

src/js/UACustomEvent.js
rlepinski: no, anything is allowed

src/js/UACustomEvent.js
```diff
@@ -39,9 +43,9 @@ class UACustomEvent {
      * Adds a property to the custom event.
      *
      * @param {string} name The property name.
-     * @param {string|number|boolean|string[]} value The property value.
+     * @param {string|number|boolean|JsonValue} value The property value.
```
marc-scig, line 46: if you're going to use JsonValue you don't need the other types in this union, as it's inclusive of them

src/js/index.d.ts
```diff
@@ -19,11 +23,11 @@ declare class UACustomEvent {
    * Adds a property to the custom event.
    *
    * @param {string} name The property name.
-   * @param {string|number|boolean|string[]} value The property value.
+   * @param {string|number|boolean|JsonValue} value The property value.
```
marc-scig, line 26: same comment here as above, re: unneeded union types

PR #276, opened 2020-05-29
Add missing React import in TS definition

barsumek:
### What do these changes do?
Typescript definition file declares such component:
`UAMessageView extends React.Component`
but `React` type is not imported. Added it.

Missing import causes an error for Typescript compiler.

Typescript compiler passes correctly.

No need.

Would be nice to e.g. add a CI step to ensure that TS definition is valid in the future.

### DIFF
--- a/src/js/index.d.ts
+++ b/src/js/index.d.ts
@@ -1,3 +1,4 @@
+import React from "react";
 import { EmitterSubscription } from "react-native";
 
 declare type JsonValue = string | number | boolean | null | JsonMap | JsonArray;


marc-scig:
@barsumek Thanks for this. I think it slipped our attention because the typescript app we were using to test it already imports React on its own. There are plans to at least add some TS tests sometime soon, but we're also considering doing a rewrite in TS.

Can you fill out this [contribution form](https://docs.google.com/forms/d/e/1FAIpQLScErfiz-fXSPpVZ9r8Di2Tr2xDFxt5MgzUel0__9vqUgvko7Q/viewform) if you haven't already?

PR #280, opened 2020-06-16
[MOBILE-1646] Add location module

oristanovic:
### What do these changes do?
Move the location code into a separate module.

Makes the integration easier for customers.

Checked that the new module was added and tested using the sample app.


rlepinski:
We also want to move location methods to the module. It will be a breaking change but we will do a major release.

urbanairship-location-react-native/README.md
```diff
@@ -0,0 +1,45 @@
...
+npm install urbanairship-location-react-native --save
+```
+
+## iOS Setup
+
+1) Install pods
+```
+cd ios && pod install
+```
+
```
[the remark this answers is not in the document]
oristanovic: I don't see any Android-specific setup that's required.

example/package.json
```diff
@@ -1,24 +1,25 @@
...
-    "react-navigation-stack": "^2.3.7",
-    "react-navigation-tabs": "^2.5.5",
-    "urbanairship-accengage-react-native": "*",
-    "urbanairship-react-native": "*"
-  },
-  "scripts": {
-    "run:android": "react-native run-android",
-    "run:ios": "react-native run-ios"
-  }
+   "private":true,
```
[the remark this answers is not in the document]
oristanovic: nope, fixed.

urbanairship-location-react-native/android/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.urbanairship.reactnative.location">
```
rlepinski, line 2: I am pretty sure we need to add the location permission to the manifest, i dont think our module includes it by default but it should start.

urbanairship-location-react-native/scripts/sync_version.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -ex
```
rlepinski, line 1: could this be shared across all modules?

urbanairship-location-react-native/scripts/sync_version.sh
```diff
@@ -0,0 +1,14 @@
...
+VERSION=$1
+
+if [ -z "$1" ]
+  then
+    echo "No version supplied"
+    exit
+fi
+
+# Update AirshipLocationReactModuleVersion.m for current install
+sed -i '' "s/\(moduleVersionString *= *@\)\".*\"/\1\"$VERSION\"/g" ios/AirshipLocationReactModuleVersion.m
```
rlepinski, line 14: we actually dont need to bother with this version

PR #281, opened 2020-06-17
MOBILE-1644 TypeScript rewrite

marc-scig:
This is a rewrite of the urbanairship-react-native module in TypeScript. I didn't modify the example except for minor cleanup, which we can do at some point later if we want. It's also probably good to demonstrate that regular react native apps can use the module with no problems.

There are a lot of changes here, partly because I reformatted several files in VS to standardize whitespace usage. The TypeScript versions of our module and test files are _almost_ similar to Flow, but differ in subtle ways. Since this is going to be part of a major update, I renamed some of the classes/typings and also standardized the module semantics to avoid exporting defaults, instead using named exports across the board. For more info on why default exports are considered problematic, see [this chapter](https://basarat.gitbook.io/typescript/main-1/defaultisbad) of TypeScript Deep Dive. I think we should do this for the accengage and location modules as well in a follow-up ticket.

I rewrote and got the unit tests passing, and have also gotten the example app running on iOS and Android. I think it's looking pretty healthy, and all the basic functionality in the app seems fine but I haven't exhaustively covered every method. Please review carefully, and if anyone wants to help manually test I'd appreciate it.

A few notes about compilation and how this affects our workflow – this uses [Bob](https://github.com/react-native-community/bob) to manage builds, and hooks into the yarn prepare step which is automatically run after installing packages. It's using Babel to compile to JS, and `tsc` to compile the typescript definitions. Those files all end up in the `lib/commonjs` and `lib/typescript` subdirectories, and the module index now points there.

I used the `ts-jest` preset as well, so that the unit tests also go through the `tsc` compiler instead of the default, which is Babel. An important detail about Babel is that it doesn't do any type-checking, rather it only strips type information. This makes sure that the the tests are type-checked before being run, adding another layer of safety.


urbanairship-react-native/package.json
```diff
@@ -15,29 +14,57 @@
   ],
   "devDependencies": {
     "@babel/core": "7.5.4",
-    "@babel/plugin-transform-flow-strip-types": "^7.4.4",
-    "@babel/runtime": "7.5.4",
+    "@react-native-community/bob": "^0.15.1",
```
marc-scig, line 17: Something to be aware of for future typescript/react work – this is a recent update to bob from a few days ago that fixes a bug that would make it return with a successful exit code when tsc reported syntax errors. We should make sure we don't use versions prior to this.

urbanairship-react-native/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "jsx": "react-native",
+    "lib": ["es2015"],
+    "moduleResolution": "node",
+    "strict": true,
+    "target": "es5",
+    "noImplicitAny": true
```
marc-scig, line 9: This is an important setting that's false by default. Otherwise variables whose type can't be inferred will be treated as `any`, which is convenient for translating vanilla JS to TS, but covers up issues code that can (and should) be made more type-safe.

urbanairship-react-native/package.json
```diff
@@ -15,29 +14,57 @@
...
+  "jest": {
+    "preset": "ts-jest",
+    "modulePathIgnorePatterns": ["lib/"]
+  },
+  "scripts": {
+    "prepare": "bob build",
+    "test": "jest --verbose",
+    "generate-docs": "typedoc --out docs --exclude js/__tests__ js/"
+  },
+  "types": "lib/typescript/index.d.ts",
```
marc-scig, line 63: index.d.ts is now autogenerated, so in theory people this should help keep it from getting out of sync.

.github/workflows/release.yml
```diff
@@ -66,6 +68,7 @@ jobs:
       - name: Publish library
         run: |
           cd urbanairship-react-native/
+          yarn prepare
```
rlepinski, line 71: pretty sure prepare runs before publish

urbanairship-react-native/android/gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
+@rem ##########################################################################
```
rlepinski, line 2: What does this buy us having it in the plugin folder?

urbanairship-react-native/js/UAEventEmitter.ts
```diff
@@ -1,33 +1,38 @@
...
-  addListener(eventType: string, listener: Function, context: ?Object): EmitterSubscription {
+  addListener(eventType: string, listener: (...args: any[]) => any, context?: object | null | undefined): EmitterSubscription {
     if (Platform.OS === 'android') {
       UrbanAirshipModule.addAndroidListener(eventType);
     }
     return super.addListener(eventType, listener, context);
   }
 
-  removeAllListeners(eventType: string) {
+  removeAllListeners(eventType?: string) {
```
rlepinski, line 29: why would event type be nullabale?

.github/workflows/release.yml
marc-scig: I wasn't sure about that, but you're probably right

urbanairship-react-native/android/gradlew.bat
marc-scig: hmm, this must have been a weird merge artifact

urbanairship-react-native/js/UAEventEmitter.ts
marc-scig: so that you can call `removeAllListeners()` to remove all listeners for all event types. its typed that way in the base class: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L215

.github/workflows/release.yml
marc-scig: looks like the answer is yes but fyi https://github.com/yarnpkg/yarn/issues/4634

urbanairship-react-native/js/__tests__/CustomEvent.test.ts
```diff
@@ -1,11 +1,11 @@
 /* Copyright Airship and Contributors */
 
-const UACustomEvent =  require('../UACustomEvent')
+import { CustomEvent } from '../CustomEvent';
 
-describe("UACustomEvent Tests", () => {
+describe("CustomEvent Tests", () => {
```
rlepinski, line 5: Breaking change right? We will need to note these for a release

marc-scig: Yeah, this one and (UA)MessageView are the only renamed classes. The typescript definitions will be different, due to minor interface changes and renaming, but the overall API is the same except for removed deprecations.

urbanairship-react-native/js/UAEventEmitter.ts
rlepinski: What about https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-native/index.d.ts#L9256 ?

rlepinski: I am using `"@types/react-native": "^0.62.13"` and get a ts build error if i try to call it with a nullable

urbanairship-react-native/js/AttributeEditor.ts
```diff
@@ -0,0 +1,100 @@
...
+        attributeType = "date";
+        attributeValue = value.getTime();
+      } else {
+        throw "Unsupported attribute type: " + typeof attributeValue;
+      }
+    } else {
+      // No boolean attribute type. Convert value to string.
+      attributeValue = value.toString();
+      attributeType = "string";
+    }
```
BrianBatchelder, line 75: I think this if would read better if all of the tests were positive, i.e.
```suggestion
    if (typeof value === "string") {
      attributeType = "string";
      attributeValue = value;
    } else if (typeof attributeValue === "number") {
      attributeType = "number";
      attributeValue = value;
    } else if (typeof value === "boolean") {
      // No boolean attribute type. Convert value to string.
      attributeType = "string";
      attributeValue = value.toString();
    } else if (value instanceof Date) {
      // JavaScript's date type doesn't pass through the JS to native bridge.
      // Dates are instead serialized as milliseconds since epoch.
      attributeType = "date";
      attributeValue = value.getTime();
    } else {
      throw "Unsupported attribute type: " + typeof attributeValue;
    }
```

urbanairship-react-native/js/UAEventEmitter.ts
```diff
@@ -1,33 +1,38 @@
 /* Copyright Airship and Contributors */
```
BrianBatchelder, line 1: Any reason this class isn't renamed "EventEmitter", like the other classes?

urbanairship-react-native/package.json
BrianBatchelder: Looks like we are not storing it after it is generated. What happens in a fresh checkout?

marc-scig: `lib/` is .gitignore, so the idea is that the directory and its contents will be auto-generated locally instead of being checked in to version control. Think of `lib/` more like an obj or bin folder and not source code.

urbanairship-react-native/js/AttributeEditor.ts
marc-scig: Your example actually won't compile because attributeValue hasn't been assigned to in the final else clause. That's why boolean is special cased (underlyingly, it's not a valid attribute type). I can rewrite it to use a positive test for boolean-ness though.

urbanairship-react-native/js/AttributeEditor.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const operation = { "action": "set", "value": attributeValue, "key": name, type: attributeType};
```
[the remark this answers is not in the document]
marc-scig: AttributeOperation is an interface, not a class. Interfaces in TypeScript are similar to those in Go – any object with those keys and value types already implements it implicitly. This also helps for compatibility with plain JavaScript since you can just keep passing an object/map like before and not have to worry about constructors/classes/etc.

urbanairship-react-native/js/MessageView.tsx
```diff
@@ -0,0 +1,143 @@
...
+ */
+export interface MessageClosedEvent {
+  /**
+   * The message ID.
+   */
+  messageId: string
+}
+
+/**
+ * UAMessageView props
```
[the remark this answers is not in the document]
marc-scig: Ah yes, thank you. I renamed this kind of late in the game.

urbanairship-react-native/js/UAEventEmitter.ts
marc-scig: There's already an EventEmitter in the react native typings, and this one derives from it. I thought about changing the name to something else, but I couldn't come up with something more descriptive so I left it alone.

PR #286, opened 2020-07-09
[fix-publishing] publish new modules

marc-scig:
Adds the new modules to our publishing workflow, and also removes an unneeded airshipVersion re-definition that was out of date.


urbanairship-location-react-native/android/build.gradle
```diff
@@ -2,8 +2,6 @@ apply plugin: 'com.android.library'
 apply from: "../../common-build.gradle"
 
 dependencies {
-    def airshipVersion = "13.1.2"
```
crow, line 5: where is this defined now?

PR #288, opened 2020-07-15
[fix-npm-pack] remove file specifiers

crow:
We should probably stick with .npmignore where necessary and skip using the files property


rlepinski:
Mostly works but the common build.gradle is messed up. For now we should remove it and inline it in each module.

PR #292, opened 2020-07-29
[MOBILE 1797] Edited removeListener desciption, added removeAllListeners and removeSubscription methods

Jahdeh:
### What do these changes do?
These changes change the `removeListener` method so that it takes an Event Emitter subscription to pass onto `removeSubscription` in `UAEventEmitter.ts`, which should remove the listener. I also added UAEventEmitter's `removeAllListeners` to `UrbanAirship.ts` as it wasn't previously in there, and should let users remove all event listeners of a certain type if they so choose.

`removeListener` doesn't work in its current state. I believe it's supposed to be inherited from React Native itself, but no combination of parameters actually seem to remove the listener (I've tried passing the event type as the first argument, and the emitter subscription as the second argument). From [this stack overflow question](https://stackoverflow.com/questions/36886628/how-do-you-remove-a-listener-from-react-natives-eventemitter-instance/36901160), it does seem that the best way to remove a listener is by removing the emitter through `removeSubscription`. As for why the react native removeListener doesn't work, I'm really not quite sure. Maybe it doesn't know what individual listener to remove by the event type?

As for adding `removeAllListeners`, this was already in `UAEventEmitter`, but wasn't being used, so I added it.

To verify, I made the changes in the react-native-module, and installed that to my react native project and created buttons calling the methods in UrbanAirship.ts to add a listener, remove a listener, and remove all listeners.

1. Pull this branch
2. `npm link`, then `npm pack` in react-native-module
3. `npm install <tarball created from npm pack>`
4. In your test app, add some event listeners (can be any type, though I mainly tested with deep link).
5. Call `removeAllListeners`, all event listeners should be removed.
6. Add some more event listeners, and make sure to store the emitter subscription returned from `addListener`.
7. Call `removeListener` passing in emitter subscription.

Let me know if you guys think this is the right approach, I want to make sure it works with what y'all have planned for the react native library.

## EDIT

After talking with Marc, it was determined that the `removeListener` method was working correctly, though is rather confusing to use. The second parameter expects a reference to the original listener function, which many devs wouldn't think to create before passing it into `addListener`. Because of that, I've edited the description of `addListener` to make that a bit more clear.

Also, I think its still beneficial to add `addSubscription` and `removeAllListeners`, so I've gone ahead and done that as well.

### DIFF
--- a/urbanairship-react-native/js/UrbanAirship.ts
+++ b/urbanairship-react-native/js/UrbanAirship.ts
@@ -653,7 +653,7 @@ export class UrbanAirship {
    * EventType.InboxUpdated, or EventType.ShowInbox.
    */
   static removeAllListeners(eventType: EventType) {
-    EventEmitter.removeAllListeners(eventType);
+    EventEmitter.removeAllListeners(convertEventEnum(eventType));
   }
 
   /**
--- a/urbanairship-react-native/js/UrbanAirship.ts
+++ b/urbanairship-react-native/js/UrbanAirship.ts
@@ -639,7 +639,7 @@ export class UrbanAirship {
   /**
    * Removes a listener for an Urban Airship event.
    *
-   * @param listener The event listner (also an emitter subscription).
+   * @param listener The event listener (an emitter subscription).
    */
   static removeListener(listener: EmitterSubscription) {
     EventEmitter.removeSubscription(listener);
--- a/urbanairship-react-native/js/UrbanAirship.ts
+++ b/urbanairship-react-native/js/UrbanAirship.ts
@@ -629,7 +629,7 @@ export class UrbanAirship {
    * @param eventType The event type. Either EventType.NotificationResponse, EventType.pushReceived,
    * EventType.Register, EventType.Reistration, EventType.DeepLink, EventType.NotificationOptInStatus,
    * EventType.InboxUpdated, or EventType.ShowInbox.
-   * @param listener The event listner.
+   * @param listener The event listener.
    * @return An emitter subscription.
    */
   static addListener(eventType: EventType, listener: (...args: any[]) => any): EmitterSubscription {
@@ -642,7 +642,7 @@ export class UrbanAirship {
    * @param eventType The event type. Either EventType.NotificationResponse, EventType.pushReceived,
    * EventType.Register, EventType.Reistration, EventType.DeepLink, EventType.NotificationOptInStatus,
    * EventType.InboxUpdated, or EventType.ShowInbox.
-   * @param listener The event listner. Should be a reference to the function passed into addListener.
+   * @param listener The event listener. Should be a reference to the function passed into addListener.
    */
   static removeListener(eventType: EventType, listener: (...args: any[]) => any) {
     EventEmitter.removeListener(convertEventEnum(eventType), listener);
--- a/urbanairship-react-native/js/UrbanAirship.ts
+++ b/urbanairship-react-native/js/UrbanAirship.ts
@@ -636,13 +636,25 @@ export class UrbanAirship {
     return EventEmitter.addListener(convertEventEnum(eventType), listener);
   }
 
-  /**
+   /**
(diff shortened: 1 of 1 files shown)


urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -639,13 +639,21 @@ export class UrbanAirship {
   /**
    * Removes a listener for an Urban Airship event.
    *
+   * @param listener The event listener (an emitter subscription).
+   */
+  static removeListener(listener: EmitterSubscription) {
+    EventEmitter.removeSubscription(listener);
```
BrianBatchelder, line 645: According to the [stack overflow question](https://stackoverflow.com/questions/36886628/how-do-you-remove-a-listener-from-react-natives-eventemitter-instance/36901160) you referenced, the listener is removed by calling `remove()` on the subscription, or in this case:
```
listener.remove();
```
It looks like both ways work, but I'm curious why did you not do it that way?

Jahdeh: `removeSubscription` in UAEventEmitter had some additional code in it, I figured that was more important in terms of Airship.

```
removeSubscription(subscription) {
    if (_reactNative.Platform.OS === 'android') {
      UrbanAirshipModule.removeAndroidListeners(1);
    }

    super.removeSubscription(subscription);
  }
```

BrianBatchelder: Ah. Good catch. I didn't realize `EventEmitter` was a `UAEventEmitter`.

urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -642,12 +642,32 @@ export class UrbanAirship {
...
   static removeListener(eventType: EventType, listener: (...args: any[]) => any) {
     EventEmitter.removeListener(convertEventEnum(eventType), listener);
   }
 
+  /**
+   * Removes a subscription for an Urban Airship event.
+   *
+   * @param subscription The emitter subscription, returned from addListener.
+   */
+  static removeSubscription(subscription: EmitterSubscription) {
```
marc-scig, line 656: I'm actually not sure we need this after all, since you can call `remove()` on the subscription itself which does the exact same thing. https://github.com/facebook/react-native/blob/master/Libraries/vendor/emitter/_EmitterSubscription.js#L53

That's a pretty typical API for things like subscriptions and disposables. The removeAllListeners method is still useful though since it doesn't require hanging onto any state.

Jahdeh: Me and Brian discussed the first part actually (It should be in our conversation up above)! `removeSubscription` was already created in UAEventEmitter, and there's some pre-existing code in there that I figured was important to Airship, so I thought it may be worth adding it to Urbanairship.ts. If that extra code in the method isn't needed, then we can remove it and have folks use remove().

What do you think?

urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -642,12 +642,32 @@ export class UrbanAirship {
    * @param eventType The event type. Either EventType.NotificationResponse, EventType.pushReceived,
    * EventType.Register, EventType.Reistration, EventType.DeepLink, EventType.NotificationOptInStatus,
    * EventType.InboxUpdated, or EventType.ShowInbox.
-   * @param listener The event listner.
+   * @param listener The event listner. Should be a reference to the function passed into addListener.
```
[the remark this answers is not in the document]
Jahdeh: Fixed! Thanks!

urbanairship-react-native/js/UrbanAirship.ts
marc-scig: Sorry to leave you hanging there – after more discussion I think we should take out `removeSubscription` and call it good. Since the subscription itself already has a cancel method it's arguably redundant, and rather than giving a bunch of different options we'd be better off just documenting the plugin better.

PR #293, opened 2020-08-13
[MOBILE-1720] Publish library docs for new react native modules

BrianBatchelder:
### What do these changes do?
Create, bundle and publish library docs for all react native modules.

We added new modules, and devs need to be able to see their APIs.

Built and browsed the docs. Will try uploading once people have reviewed the new format.

I've attached the tar ([Documentation.tar.gz](https://github.com/urbanairship/react-native-module/files/5069980/Documentation.tar.gz)) that is generated by the upload script. Please download and browse.

- We should also change our react platform docs to have a link to each module's API doc, as we do for the [iOS platform docs](https://docs.airship.com/platform/ios/getting-started/).
 - I deleted `documentation.yml`, as I think it was used by the previous documentation tool so is no longer needed.


documentation/index-for-docs.html
```diff
@@ -0,0 +1,17 @@
+<html>
```
BrianBatchelder, line 1: This is a fairly ugly index.html I added just in case someone happens to land at the top level of the docs. But I think we should add direct links to each doc from our platform docs, so it will be highly unlikely anyone will land here.

urbanairship-accengage-react-native/package.json
```diff
@@ -40,7 +40,7 @@
   "scripts": {
     "prepare": "bob build",
     "test": "jest --verbose",
-    "generate-docs": "typedoc --out docs --exclude js/__tests__ js/"
+    "generate-docs": "typedoc --out docs"
```
BrianBatchelder, line 43: The exclude of `js/__tests__` and the include of `js/` are covered by the `tsconfig.json` file, so they don't need to be here.

urbanairship-location-react-native/package.json
```diff
@@ -40,7 +40,7 @@
   "scripts": {
     "prepare": "bob build",
     "test": "jest --verbose",
-    "generate-docs": "typedoc --out docs --exclude js/__tests__ js/"
+    "generate-docs": "typedoc --out docs --exclude **/index.ts"
```
BrianBatchelder, line 43: I'm excluding `index.ts` because it just exports the other methods and properties, which cluttered up the docs.

urbanairship-react-native/tsconfig.json
```diff
@@ -8,6 +8,6 @@
     "target": "es5",
     "noImplicitAny": true
   },
-  "include": ["js/**/*.ts"],
+  "include": ["js/**/*.ts", "js/**/*.tsx"],
```
BrianBatchelder, line 11: I added the `tsx` files to pick up MessageView. One concern I have is that MessageView will now be added to the build, and maybe we didn't want it there?

rlepinski: @marc-scig I assume we want it part of the build?

PR #313, opened 2021-03-20
[MOBILE-2152] Rework events to support background events on Android

rlepinski:
This reworks the event emitter in order to take advantage of headlessJS services on Android to notify some events in the background. iOS already supported events in the background due to when the react context is constructed, but Android waits until the activity is started.

We use to do Native -> Event -> Has listener? -> Dispatch event to JS. Everything was a 1 to 1 on the app listener for a certain event on the NativeEventEmitter. With this change, we instead do this:  Native -> Event -> Notify Pending to JS. Then the JS layer will take events for any known listener from the plugin and dispatch them.

The main reason for this change is to allow the headless task to fully process the event before finishes the service. I had no way of doing that with NativeEventEmitter, the emit call did not return a promise or a way to set callback. 

There is a minor breaking API change (EmitterSubscription -> Subscription) and possible breaking behavior changes if the app is doing UI work on Android from a background event. The sample would crash because we never cleaned up the event emitters on unmount.

I tested it on both iOS and Android. I need to test content-available pushes still on iOS to verify push-received events in the background. To make this work on Android you need to send a high priority FCM message.

This is to support - https://github.com/Expensify/Expensify.cash/issues/1097


rlepinski:
I modified the UAEventEmitter to make it testable without having to deal with any of the react-native objects

urbanairship-react-native/js/UAEventEmitter.ts
```diff
@@ -2,42 +2,105 @@
...
+class DefaultAirshipEventBridge extends AirshipEventBridge {
+  nativeModule = NativeModules.UrbanAirshipReactModule;
+  eventEmitter = new NativeEventEmitter(this.nativeModule);
+
+  constructor(dispatchEventsCallback: DispatchEventsCallback) {
+    super(dispatchEventsCallback);
+
     if (Platform.OS === 'android') {
-      UrbanAirshipModule.addAndroidListener(eventType);
+      AppRegistry.registerHeadlessTask('AirshipAndroidBackgroundEventTask', () => {
```
marc-scig, line 33: should this function also be marked async?

urbanairship-react-native/js/UAEventEmitter.ts
```diff
@@ -2,42 +2,105 @@
...
+}
+
+/**
+ * SDK event emitter.
+ *
+ * @hidden
+ */
+export class UAEventEmitter {
+  airshipEventBridge: AirshipEventBridge;
+  listeners: Map<string, Array<(...args: any[]) => any>>;
```
marc-scig, line 59: I kind of feel like if we're moving away from the underlying event emitter we might be able to get away with a more explicit typing for our listener functions than just `(...args: any[]) => any`, but maybe we can save that for another time

marc-scig (approved):
I had a non-blocking comment and a question about whether a thing should be marked async, otherwise looks 💯

urbanairship-react-native/js/UAEventEmitter.ts
rlepinski: The task provider does not seem to take an async callback

jyaganeh (approved):

PR #317, opened 2021-03-25
[MOBILE-2136] release 11.0

marc-scig:
This release technically also supports React Native 0.6.4/React 17 but we don't need to call that out. Minimum requirements remain the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # React Native Module Changelog
 
+## Version 11.0.0 - March 25, 2021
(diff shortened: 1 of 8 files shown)


Ulrico972 (approved):
Looks good 👍 .
Although, I don't know if that's used anymore but you should also update the version string in `urbanairship-react-native/ios/UARCTModule/UARCTModuleVersion.m`, `urbanairship-location-react-native/ios/AirshipLocationReactModuleVersion.m` and `urbanairship-accengage-react-native/ios/AirshipAccengageReactModuleVersion.m`

jyaganeh (approved):
1️⃣ 1️⃣ 🎉

PR #332, opened 2021-07-05
[MOBILE-2359] Expose new privacy manager flags

Ulrico972:
### What do these changes do?
Updates the Airship SDKs version, expose the new privacy manager methods and remove the old data collection methods

To replace the data collection methods with the privacy manager

Tested the methods by logging the results in the sample.
The methods work fine on both Android and iOS.

We will need to add some example of use in the doc

### DIFF
--- a/urbanairship-react-native/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
+++ b/urbanairship-react-native/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
@@ -26,6 +26,7 @@ import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.bridge.WritableNativeArray;
 import com.facebook.react.bridge.WritableNativeMap;
 import com.urbanairship.PrivacyManager;
 import com.urbanairship.UAirship;
@@ -44,9 +45,12 @@ import com.urbanairship.reactnative.events.NotificationOptInEvent;
 import com.urbanairship.reactnative.events.PushReceivedEvent;
 import com.urbanairship.util.UAStringUtil;
 
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 
 import static com.urbanairship.actions.ActionResult.STATUS_ACTION_NOT_FOUND;
 import static com.urbanairship.actions.ActionResult.STATUS_COMPLETED;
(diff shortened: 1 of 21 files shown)


urbanairship-react-native/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
```diff
@@ -836,4 +850,51 @@ static void checkOptIn(Context context) {
...
+                    intFeatures[i] = PrivacyManager.FEATURE_TAGS_AND_ATTRIBUTES;
+                    break;
+                case "FEATURE_CONTACTS":
+                    intFeatures[i] = PrivacyManager.FEATURE_CONTACTS;
+                    break;
+                case "FEATURE_LOCATION":
+                    intFeatures[i] = PrivacyManager.FEATURE_LOCATION;
+                    break;
+                default:
+                    intFeatures[i] = PrivacyManager.FEATURE_ALL;
```
marc-scig, line 893: I think it would be safer to add a case for "FEATURE_ALL", and make the default label log an error. otherwise if someone is using plain javascript and they pass a string that's not in this list it would be interpreted as FEATURE_ALL, which is probably not what we want.

urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -371,47 +384,51 @@ export class UrbanAirship {
...
+  }
+
+  /**
+   * Gets a flag number representing the enabled features.
+   * 
+   * @return A promise that returns the enabled features as a flag `number`.
    */
-  static setDataCollectionEnabled(enabled: boolean) {
-    UrbanAirshipModule.setDataCollectionEnabled(enabled);
+  static getEnabledFeatures(): Promise<number> {
```
marc-scig, line 403: should getEnabledFeatures return a `Promise<Feature[]>`?

urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -525,4 +528,33 @@ - (UAAttributeMutations *)mutationsWithOperations:(NSArray *)operations {
...
+        } else if ([feature isEqualToString:@"FEATURE_ANALYTICS"]) {
+            convertedFeatures |= UAFeaturesAnalytics;
+        } else if ([feature isEqualToString:@"FEATURE_TAGS_AND_ATTRIBUTES"]) {
+            convertedFeatures |= UAFeaturesTagsAndAttributes;
+        } else if ([feature isEqualToString:@"FEATURE_CONTACTS"]) {
+            convertedFeatures |= UAFeaturesContacts;
+        } else if ([feature isEqualToString:@"FEATURE_LOCATION"]) {
+            convertedFeatures |= UAFeaturesLocation;
+        } else {
+            convertedFeatures |= UAFeaturesAll;
```
marc-scig, line 554: same observation here as in the android implementation

urbanairship-react-native/android/src/main/java/com/urbanairship/reactnative/UrbanAirshipReactModule.java
Ulrico972: Good catch, didn't think about it. Thank you very much 👍

urbanairship-react-native/js/UrbanAirship.ts
Ulrico972: Actually I was wondering that too. But since the native SDKs was returning int values I let it like this.
But I think you're right, an array of Feature should be better for the customers.
I'll make the change then.

urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -87,12 +87,29 @@ - (instancetype)init{
...
+RCT_EXPORT_METHOD(setEnabledFeatures:(NSArray *) features) {
+    [UAirship shared].privacyManager.enabledFeatures = [self parseStringFeatures:features];
 }
 
-RCT_EXPORT_METHOD(setPushTokenRegistrationEnabled:(BOOL)enabled) {
-    [[UAirship push] setPushTokenRegistrationEnabled:enabled];
+RCT_REMAP_METHOD(getEnabledFeatures,
+                 getEnabledFeatures_resolver:(RCTPromiseResolveBlock)resolve
+                 rejecter:(RCTPromiseRejectBlock)reject) {
+    resolve(@([UAirship shared].privacyManager.enabledFeatures));
```
rlepinski, line 97: I think you need to convert them back to strings right?

Ulrico972: Right 👍

urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -87,29 +89,81 @@ - (instancetype)init{
...
+                 rejecter:(RCTPromiseRejectBlock)reject) {
+   
+    if ([self isValidFeature:features]) {
+        [UAirship shared].privacyManager.enabledFeatures = [self stringToFeature:features];
+        resolve(@(YES));
+    } else {
+        NSString *code = [NSString stringWithFormat:UARCTStatusInvalidFeature];
+        NSString *errorMessage = [NSString stringWithFormat:UARCTErrorDescriptionInvalidFeature];
+        NSError *error =  [NSError errorWithDomain:UARCTErrorDomain
+                                              code:2
```
marc-scig, line 104: It would be nice to map this number to a constant, like the other errors

urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -390,36 +393,39 @@ export class UrbanAirship {
...
 
   /**
-   * Gets a flag number representing the enabled features.
+   * Gets a String array with the enabled features.
    * 
-   * @return A promise that returns the enabled features as a flag `number`.
+   * @return A promise that returns the enabled features as a String array.
    */
-  static getEnabledFeatures(): Promise<number> {
+  static getEnabledFeatures(): Promise<String[]> {
```
marc-scig, line 407: Can we make this return an array of Feature?

rlepinski (approved):
After addressing marcs comments

marc-scig (approved):

PR #337, opened 2021-07-15
fix: Add missing prop type messageId for MessageView

lukebars:
### What do these changes do?
Adds missing type definition for prop messageId for MessageView


rlepinski:
Thank you!

rlepinski (approved):

marc-scig (approved):

PR #338, opened 2021-07-15
[MOBILE-2418] fix example README

marc-scig:
We still had a reference to `npm install` in the example README, which is confusing and results in errors. The example is connected to our yarn workspace, and so you have to use yarn to build the modules.


jyaganeh (approved):
🧶

rlepinski (approved):

PR #339, opened 2021-07-19
Expose `addTags` and `removeTags` methods

ds8k:
### What do these changes do?
These changes expose the `addTags` and `removeTags` methods that both the iOS and Android SDK support

The RN library calls `updateRegistration` after each `addTag` - this is problematic when onboarding new users where bulk subscriptions are necessary.

These changes are largely copy-paste and making a change from `addTag` to `addTags`. I haven't built them to verify just yet


whileloup:
I've bumped this internally for you @ds8k

ds8k:
> I've bumped this internally for you @ds8k 

Cool. Pretty sure I've got some typing issues in here, and since we're going with tag groups there's no rush on it.

rlepinski:
Thanks for the PR. This looks good, but we might want to expose an editor interface to encourage combining all tag changes in a single update. I know this is low priority for you now, but are you wanting to make that change?

PR #343, opened 2021-08-03
[MOBILE-2420] Fix message center display

oristanovic:
Fixes the displayMessageCenter call to display the OOTB MC.
Verified that it displays well on the test app.


urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -351,13 +351,13 @@ - (instancetype)init{
 
 RCT_EXPORT_METHOD(displayMessageCenter) {
     dispatch_async(dispatch_get_main_queue(), ^{
-        [[UAMessageCenter shared] display];
```
rlepinski, line 354: This should still be display, what we need to fix is the delegate that we set needs to check if autoLaunchMessageCenter is enabled, and if so show the default UI

PR #352, opened 2021-09-03
[MOBILE-2482] isUserNotificationsOptedIn discrepancy between iOS and Android in React Native SDK Issue

Apekka:
### What do these changes do?
 - Update the iOS method isUserNotificationsOptedIn to use the UAPush method. 

- Add a new method isSystemNotificationsEnabledForApp that checks for the notification manager on android and authorized types on iOS.

- Add another method that for Android getNotificationChannelStatus(channel) that returns enabled, disabled, and unknown where unknown is returned if it's not yet created.

Because this presents a problem for Axios React Native Implementation noted as the unified call essentially returns a different result between iOS and Android. Axios has built logic into their app that requires App-Level notification opt-in information that we currently do not expose for Android. 

Running with the sample app and testing the methods

### DIFF
--- a/urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
+++ b/urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
@@ -198,7 +198,6 @@ RCT_REMAP_METHOD(isUserNotificationsOptedIn,
             UA_LTRACE(@"Opted out: push is disabled");
             optedIn = NO;
         }
-
         resolve(@(optedIn));
 }
 
--- a/urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
+++ b/urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
@@ -193,6 +193,11 @@ RCT_REMAP_METHOD(isUserNotificationsOptedIn,
             UA_LTRACE(@"Opted out: no authorized notification settings");
             optedIn = NO;
         }
+    
+        if (![[UAirship shared].privacyManager isEnabled:UAFeaturesPush]) {
+            UA_LTRACE(@"Opted out: push is disabled");
+            optedIn = NO;
+        }
 
         resolve(@(optedIn));
 }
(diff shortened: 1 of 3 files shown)


urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -508,6 +508,30 @@ export class UrbanAirship {
...
+  /**
+   * Gets the status of the specified Notification Channel.
+   * This method is only supported on Android. iOS will no-op.
+   *
+   * @param channel The channel's name.
+   * @return A promise with the result.
+   */
+  static getNotificationChannelStatus(channel: string): Promise<string> {
+    if (Platform.OS != 'android') {
+      console.log("This feature is not supported on this platform.");
```
rlepinski, line 530: can you resolve the promise with an error?

urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -178,14 +178,20 @@ - (instancetype)init{
 RCT_REMAP_METHOD(isUserNotificationsOptedIn,
                  isUserNotificationsOptedIn_resolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject) {
+    BOOL optedIn = [[UAirship push] userPushNotificationsEnabled];
```
rlepinski, line 181: you actually need to recreate the logic in this method - https://github.com/urbanairship/ios-library/blob/main/Airship/AirshipCore/Source/Internal/UAPush.m#L652

Apekka: I don't understand why I wasn't able to get "isRegisteredForRemoteNotifications" and the privacyManager from UAPush, so I haven't included them in the logic

marc-scig: I'm not so worried about isRegisteredForRemoteNotifications, but unless this includes privacy manager the method doesn't mean the same thing it does on Android. I believe you can get the shared privacy manager from UAirship

urbanairship-react-native/ios/UARCTModule/UrbanAirshipReactModule.m
```diff
@@ -178,14 +178,40 @@ - (instancetype)init{
...
+            UA_LTRACE(@"Opted out: no authorized notification settings");
+            optedIn = NO;
+        }
+    
+        if (![[UAirship shared].privacyManager isEnabled:UAFeaturesPush]) {
+            UA_LTRACE(@"Opted out: push is disabled");
+            optedIn = NO;
+        }
+
+        resolve(@(optedIn));
```
rlepinski, line 202: looks like whitespace is off here

rlepinski (approved):

Ulrico972 (approved):
Just the whitespaces that ryan mentioned

PR #389, opened 2022-01-13
[GHI-382] Fix iOS behavior with auto launch message center

Ulrico972:
### What do these changes do?
Fix iOS behavior with auto launch message center

We should not show the modal on displayMessage(messageId) when the auto launch message center is false

Manual tests

I'm not sure what to do to show the message in the custom message center when auto launch is false. Right now it doesn't seem to work


Ulrico972 (commented):
👍

oristanovic (approved):

PR #408, opened 2022-03-05
Release 14.0.0

rlepinski:
Updated to latest SDK versions, returns full preference config using some internal SDK apis (with shared types from web SDK), add subscription editor of contact, and deprecates tag and attribute methods.

### DIFF
--- a/urbanairship-react-native/js/UrbanAirship.ts
+++ b/urbanairship-react-native/js/UrbanAirship.ts
@@ -756,10 +756,20 @@ export class UrbanAirship {
    * Creates an editor to modify the named user tag groups.
    *
    * @return A tag group editor instance.
+   * @deprecated Replaced by {@link editContactTagGroups()}.
    */
   static editNamedUserTagGroups(): TagGroupEditor {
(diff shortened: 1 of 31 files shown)


urbanairship-react-native/js/UrbanAirship.ts
```diff
@@ -734,21 +745,31 @@ export class UrbanAirship {
   /**
    * Gets the subscription lists.
    *
-   * @param types The types of subscription lists. Defaults to an [`channel`].
+   * @param types The types of subscription lists. Values: `channel`, `contact` (default: [`channel`]).
    * @return A promise with the result.
    */
-   static getSubscriptionLists(types?: [SubscriptionListType]): Promise<SubscriptionLists> {
+   static getSubscriptionLists(types?: [...SubscriptionListType[]]): Promise<SubscriptionLists> {
```
rlepinski, line 751: We need to make sure the edit, get, and get preference all use the same type. I think its lower case everywhere

Ulrico972 (approved):

Issue #406, opened 2022-02-24
Question about running a app on Android emulator without a valid development app key


cchristman-shell:
Hello,

I have a question about the key for urbanairship. We don't have a valid "development app key" for User Notifications in a test application.  We are trying to test the app in an Android simulator (Pixel 2 API 30), Android Studio logs this error at run time: "java.lang.IllegalArgumentException: AirshipConfigOptions: appKey is not a valid development app key", and the app keeps stopping.  

E/AndroidRuntime: FATAL EXCEPTION: UrbanAirship#1
    Process: com.shell, PID: 7383
    java.lang.IllegalArgumentException: AirshipConfigOptions: appKey is not a valid development app key
        at com.urbanairship.AirshipConfigOptions.validate(AirshipConfigOptions.java:383)
        at com.urbanairship.UAirship.executeTakeOff(UAirship.java:402)
        at com.urbanairship.UAirship.access$000(UAirship.java:65)
        at com.urbanairship.UAirship$2.run(UAirship.java:381)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at com.urbanairship.util.AirshipThreadFactory$1.run(AirshipThreadFactory.java:50)
        at java.lang.Thread.run(Thread.java:923)

Do we need a valid development app key?   Is there a way to bypass the key or use an empty key that won't crash the app?

Dependencies
"urbanairship-react-native": "^9.0.0"
"react-native": "0.64.1"

Thanks,

❗For how-to inquiries involving Airship functionality or use cases, please
contact (support)[https://support.airship.com/].

# Preliminary Info

### What are the versions of any relevant development tools you are using?

# Report

### What is the expected behavior?

Expected the app to run in the Android simulator, but it keeps stopping.

rlepinski:
You have a few options:
- use `appKey` and `appSecret` instead of `productionAppKey` and `productionAppSecret` on Android in the config
- set the `inProduction` flag to true
- generate another project for debug builds

The SDK doesnt really care about what type of app its connecting to, production vs development is just a convenience. At runtime, if its a debug build we will pull the `development` creds, otherwise the `production` creds.

cchristman-shell:
[...]

PR #409, opened 2022-03-10
[MOBILE-2962] Make takeOff optional

rlepinski:
### What do these changes do?
Makes it so takeOff is not required for the plugin to load. All calls will either no-op or reject the promise.

This makes it possible to call takeOff from the ReactNative layer:
- More control over when and how takeOff is called
- Matches our other plugins
- Makes this no longer an issue - https://github.com/urbanairship/react-native-module/issues/406
- Needed for expo managed workflows

Ran sample app without takeOff being called. Verified everything loads without crashes. 

I cleaned up and fixed a bunch of issues:
- OpenPreferenceCenter and OpenChat was not wired up properly
- Saving preferences either unprefixed
- Saving preferences on Android in the default preferences instead of the plugin preferences
- Cleaned up event emitting on iOS
- Added a common preference class to make it more obvious in the future (still need to move to own suite on iOS)


oristanovic (approved):

jyaganeh (approved):

roryabraham:
This was a breaking change for us since we have a `CustomNotificationProvider` that extends `ReactNotificationProvider` and the `context` member went from `protected` -> `private`

rlepinski:
@roryabraham We can revert that change, sorry for that.

roryabraham:
No need to revert! It was a super easy fix. Thanks!

PR #416, opened 2022-04-13
Fix an issue with events not triggered

khmMouna:
### What do these changes do?
Fix an issue with event not triggered in iOS.

Testing manually with sample app.


urbanairship-react-native/ios/UARCTModule/UARCTAirshipListener.m
```diff
@@ -20,11 +20,21 @@ @interface UARCTAirshipListener()
 
 @implementation UARCTAirshipListener
 
-- (instancetype)initWithEventEmitter:(UARCTEventEmitter *)eventEmitter {
+static UARCTAirshipListener *airshipListener_;
+
++ (void)load {
```
rlepinski, line 25: Use a dispatch _once instead when accessing shared. `load` methods hurt startup performance.

PR #428, opened 2022-05-16
[MOBILE-2900] Rewrite in Kotlin

Ulrico972:
### What do these changes do?
Rewrite in Kotlin

Build and run


example/android/build.gradle
```diff
@@ -13,8 +13,9 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath('com.android.tools.build:gradle:4.1.0')
-        classpath("com.google.gms:google-services:4.2.0")
+        classpath('com.android.tools.build:gradle:4.1.3')
+        classpath("com.google.gms:google-services:4.3.10")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.31")
```
rlepinski, line 18: is this required to use our module?

Ulrico972: It seems to be.
I'll make sure we have no other option without adding things on the customer side.

jyaganeh: I was able to build successfully with the kotlin dependency removed from the example, and the following added to the top of each of the modules:

```groovy
// moved this up from lower down in the file...
import groovy.json.JsonSlurper

buildscript {
    ext {
        airshipVersion = "16.4.0"

        kotlinVersion = '1.3.11'
        androidToolsVersion = '3.2.1'
    }

    repositories {
        google()
        mavenCentral()
    }

    dependencies {
        classpath "com.android.tools.build:gradle:${project.ext.androidToolsVersion}"
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${project.ext.kotlinVersion}"
    }
}

apply plugin: 'com.android.library'
apply plugin: "kotlin-android"
// got rid of "kotlin-android-extensions", since that's deprecated now
```

This basically does the same thing as when this was set up in the sample, but now each module will set up its own build environment with the kotlin plugin before the rest of the file is evaluated.

jyaganeh (commented):
noticed a couple of mostly minor/nitpicky things that would be cool to fix up, but I'm not too worried about them either way

PR #458, opened 2022-10-05
[MOBILE-3287] Expose displayInterval

jyaganeh:
### What do these changes do?

* Fixes Android chat module version
* Adds support for configuring the IAA display interval at runtime

Feature request

- Confirmed the call made it to the native layers using the Android and iOS samples
- Ran tests


jyaganeh:
need to retarget this, so I'm gonna pull it down and put a new one up

PR #526, opened 2023-09-01
WIP release 16.1.0

rlepinski:


android/src/main/java/com/urbanairship/reactnative/AirshipModule.kt
```diff
@@ -27,6 +25,8 @@ class AirshipModule internal constructor(val context: ReactApplicationContext) :
     private var isOverrideForegroundDisplayEnabled: Boolean = false
     private val foregroundDisplayRequestMap = mutableMapOf<String, CompletableDeferred<Boolean>>()
 
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.Main) + SupervisorJob()
```
jyaganeh, line 28: Is there a lifecycle callback we can use to cancel this scope? I think we'd want it to be on destroy of the application (not the hosting Activity)

jyaganeh (approved):
lgtm overall, just one question about the `scope` in `AirshipModule`

android/src/main/java/com/urbanairship/reactnative/AirshipModule.kt
rlepinski: I think modules outlive the activity but i could be wrong. Should we just make it a static var?

Ulrico972: Hmm I think there is no way to listen for when the application itself is destroyed 🤔

Ulrico972 (approved):

PR #527, opened 2023-09-01
[MOBILE-3865] Add editTags support

Ulrico972:
### What do these changes do?
Add editTags method

To keep updated

Manual testing on the sample

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1,21 +1,24 @@
 PODS:
-  - Airship (17.0.3):
-    - Airship/Automation (= 17.0.3)
-    - Airship/Basement (= 17.0.3)
-    - Airship/Core (= 17.0.3)
-    - Airship/MessageCenter (= 17.0.3)
-    - Airship/PreferenceCenter (= 17.0.3)
-  - Airship/Automation (17.0.3):
+  - Airship (17.2.2):
+    - Airship/Automation (= 17.2.2)
+    - Airship/Basement (= 17.2.2)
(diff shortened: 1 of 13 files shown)


example/ios/Podfile.lock
```diff
@@ -315,10 +318,14 @@ PODS:
   - React-logger (0.71.1):
     - glog
   - react-native-airship (16.0.1):
-    - AirshipFrameworkProxy (= 3.0.2)
+    - AirshipFrameworkProxy (= 4.1.0)
     - React-Core
-  - react-native-safe-area-context (4.7.1):
+  - react-native-safe-area-context (4.5.0):
```
jyaganeh, line 323: Did we want to downgrade this and the couple other dependencies below? Not sure if it makes much difference either way 🤷

jyaganeh (approved):

example/ios/Podfile.lock
Ulrico972: These are dependencies with a caret operator (^):
``` json
"react-native-gesture-handler": "^2.8.0",
"react-native-safe-area-context": "^4.4.1",
"react-native-screens": "^3.18.2",
```
I'm surprised `yarn install` get those versions for me if there are higher versions... But since it's just for the Podfile.lock I think it's safe to say it's no big deal (I'll confirm that with mouna or ogy tomorrow).

Apekka (approved):

example/ios/Podfile.lock
Ulrico972: I confirmed with mouna it's weird but it's fine

PR #529, opened 2023-09-21
Fix old architecture channelEditTags method

rlepinski:
This fixes an issue with channelEditTags for the old architecture.  When an interface is changed we have to manually update the old arch spec. 

I will add a follow up ticket to update the CI to build both arch types for Android and iOS to make sure we are not missing anything


Ulrico972 (approved):
Tested the old arch and it's working fine

PR #533, opened 2023-11-09
[MOBILE-3980] Add enableChannelCreation method

Ulrico972:
### What do these changes do?
Add an enableChannelCreation method and mark it as deprecated along with the isChannelCreationDelayEnabled config.

To not block people that would still use the isChannelCreationDelayEnabled config.

Tested on iOS.
Android is working fine at APK.On my side, something with my environment is blocking for now, about CMake in the NDK...

### DIFF
--- a/ios/RTNAirship.mm
+++ b/ios/RTNAirship.mm
@@ -117,6 +117,15 @@ RCT_REMAP_METHOD(channelEditTags,
     [self handleResult:nil error:error resolve:resolve reject:reject];
 }
 
+RCT_REMAP_METHOD(channelEnableChannelCreation,
+                 channelEnableChannelCreation:(RCTPromiseResolveBlock)resolve
+                 reject:(RCTPromiseRejectBlock)reject) {
+    NSError *error;
+    [AirshipReactNative.shared channelEnableChannelCreationAndReturnError:&error];
+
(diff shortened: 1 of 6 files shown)


ios/RTNAirship.mm
```diff
@@ -117,6 +117,15 @@ + (BOOL)requiresMainQueueSetup {
     [self handleResult:nil error:error resolve:resolve reject:reject];
 }
 
+RCT_REMAP_METHOD(channelEnableChannelCreation,
+                 channelEnableChannelCreation:(RCTPromiseResolveBlock)resolve
+                 reject:(RCTPromiseRejectBlock)reject) {
+    NSError *error;
+    [AirshipReactNative.shared channelEnableChannelCreationAndReturnError:&error];
```
Ulrico972, line 124: I'm not sur about this... I think it's right to add that "AndReturnError" thing but I don't fully understand how it would work lol.
Also, I'm wondering why we still have this class in objective C when we have those methods tagged with @objc in the Swift class?

rlepinski: We need the obj-c for c++ bindings with the react plugin. This looks correct to me, thats standard objc bindings for errors

khmMouna (approved):

PR #537, opened 2023-12-07
Release 17.1.0

crow:
### What do these changes do?
Minor release that updates the iOS SDK to 17.7.0 and Android SDK to 17.6.0 and adds support for notifying the contact of a remote login.

- Updated iOS SDK to 17.7.0
- Updated Android SDK to 17.6.0
- Added `Airship.contactManager.notifyRemoteLogin()` method to notify contact of remote login

🚀

Set breakpoints on notifyRemoteLogin in iOS and Android and ensured these got called when I called it:
image
image

### DIFF
--- a/ios/AirshipReactNative.swift
+++ b/ios/AirshipReactNative.swift
@@ -367,6 +367,11 @@ public extension AirshipReactNative {
     }
 
     @objc
+    func notifyRemoteLogin() throws {
+        try AirshipProxy.shared.contact.notifyRemoteLogin()
+    }
+
+    @objc
     func contactGetNamedUserIdOrEmtpy() async throws -> String {
         return try await AirshipProxy.shared.contact.getNamedUser() ?? ""
     }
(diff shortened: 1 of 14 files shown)


ios/AirshipReactNative.swift
```diff
@@ -366,6 +366,11 @@ public extension AirshipReactNative {
         try AirshipProxy.shared.contact.reset()
     }
 
+    @objc
+    func notifyRemoteLogin() throws {
```
rlepinski, line 370: You are missing the objc++ method that needs to call through to this

src/NativeRTNAirship.ts
```diff
@@ -56,6 +56,7 @@ export interface Spec extends TurboModule {
   // Contact
   contactIdentify(namedUser: string): Promise<void>;
   contactReset(): Promise<void>;
+  contactNotifyRemoteLogin(): Promise<void>;
```
rlepinski, line 59: needs to match this

PR #541, opened 2024-01-05
Check ExtraProperties for "airshipHmsEnabled" additionally

stetbern:
### What do these changes do?
Check for the "airshipHmsEnabled" flag in ExtraProperties from build.gradle on android top level additionally to check in gradle.properties.

HMS requires "com.huawei.hms:push", which then requires the REQUEST_INSTALL_PACKAGES permission. This permission is considered as "high risk or sensitive permission" by Google (see https://support.google.com/googleplay/android-developer/answer/12085295?hl=en). 

Making the "airshipHmsEnabled" flag available in ExtraProperties of the root projects makes it easier to control wether HMS should be used or not in react native apps which are build for HarmonyOS and Android at the same time. By that we can avoid the whole "REQUEST_INSTALL_PACKAGES" mess in Playstore easier, because changing values in build.gradle files can be done with less effort.

Tested in a RN app, "airshipHmsEnabled" not set in gradle.properties but in build.gradle on android top level. HMS still worked.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -21,7 +21,7 @@ def isNewArchitectureEnabled() {
 }
 
 def isHmsEnabled() {
-    return rootProject.hasProperty("airshipHmsEnabled") && rootProject.getProperty("airshipHmsEnabled") == "true"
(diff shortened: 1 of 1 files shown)


rlepinski:
Looks good to me 👍

crow (approved):

PR #542, opened 2024-01-05
fix(android): missing annotation on feature flag method

idrakimuhamad:
### What do these changes do?
The `featureFlagManagerFlag` and `featureFlagManagerTrackInteraction` missed out the annotation to make it accessible from JS for Android native-module

Without this annotation, we can't use the `flag` and `trackInteraction` method on Android.

Added the annotation, rebuild the app, and verify that the method able to call the native method

Should fix https://github.com/urbanairship/react-native-airship/issues/540


rlepinski:
👍

crow (approved):

PR #544, opened 2024-01-17
Update react-native-airship.podspec

WoLewicki:
### What do these changes do?

They make the new arch work in this library for newest RN versions, as recommended by Meta. Similar to other PRs such as: https://github.com/react-native-cameraroll/react-native-cameraroll/pull/577 or https://github.com/react-native-webview/react-native-webview/pull/3231 which was approved by Meta member.

Without them, new arch does not work on newest RN version.

We use this code in our app running on new arch: https://github.com/Expensify/App/pull/13767/files#diff-e4eb696f711269bf62811c2d5bc2500655ae6cd589e1b84b520a68b32e7fd05f

### DIFF
--- a/react-native-airship.podspec
+++ b/react-native-airship.podspec
@@ -16,22 +16,10 @@ Pod::Spec.new do |s|
 
   s.source_files = "ios/**/*.{h,m,mm,swift}"
 
-  s.dependency "React-Core"
-
-  # Don't install the dependencies when we run `pod install` in the old architecture.
-  if ENV['RCT_NEW_ARCH_ENABLED'] == '1' then
-    s.compiler_flags = folly_compiler_flags + " -DRCT_NEW_ARCH_ENABLED=1"
-    s.pod_target_xcconfig    = {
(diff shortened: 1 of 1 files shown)


rlepinski:
@WoLewicki Does this break new arch for react versions RN < 0.73 ? Do we need to extend the if else to fallback to old new arch check or will that break things?

rlepinski:
Ok, so i think we want to do this -https://github.com/react-native-webview/react-native-webview/pull/3231/files so we can avoid breaking new arch users that are on .68 - .71

WoLewicki:
`install_modules_dependencies` method is available from RN 0.71 iirc, and there is no sense in supporting new arch on lower versions since there were so many breaking changes to it already that is it unusable there. If this methods is not found, it fallbacks to the old behavior on old arch also, so it should be backward compatible on all versions of RN on old arch, and enough versions on new arch, so I think there is no breaking change in it.

crow (approved):

PR #557, opened 2024-04-05
[MOBILE-4239] Fix Airship Actions running

Ulrico972:
### What do these changes do?
Update the actions running code to use an Action object as the React conversion can't handle dynamic type like JsonValue directly.

To handle actions with different value types.

Tested several actions making sure I test every values types possible.
New arch and old arch are working fine on Android. iOS is fine too.

That's changes the API for the customers

### DIFF
--- a/ios/AirshipReactNative.swift
+++ b/ios/AirshipReactNative.swift
@@ -328,10 +328,10 @@ public extension AirshipReactNative {
 // Actions
 @objc
 public extension AirshipReactNative {
-    func actionsRun(actionName: String, actionValue: Any?) async throws-> Any? {
+    func actionsRun(action: [String: Any]) async throws-> Any? {
         return try await AirshipProxy.shared.action.runAction(
-            actionName,
-            value: try AirshipJSON.wrap(actionValue)
+            action["_name"] as! String,
+            value: action["_value"] is NSNull ? nil : try AirshipJSON.wrap(action["_value"])
         )
     }
 }
--- a/src/Action.ts
+++ b/src/Action.ts
@@ -1,0 +1,34 @@
+/* Copyright Airship and Contributors */
+
+'use strict';
+
+import { JsonValue } from './types';
+
+/**
+ * Airship Action Object.
+ * This is used to encapsulate the Action name and the Action value.
+ */
+export class Action {
+    _name: string;
+    _value?: JsonValue;
+
+    /**
+     * Airship Action constructor.
+     *
+     * @param name The action name.
+     * @param value The action value.
+     */
+    constructor(name: string, value?: JsonValue) {
+      this._name = name;
+      this._value = value;
+    }
+
+    /**
+     * Sets the action value.
+     *
+     * @param value The action value.
+     */
+    setValue(value?: JsonValue) {
+        this._value = value;
+    }
+}
--- a/src/AirshipActions.ts
+++ b/src/AirshipActions.ts
@@ -1,4 +1,5 @@
 import { JsonValue } from './types';
+import { Action } from './Action';
 
 /**
  * Airship actions.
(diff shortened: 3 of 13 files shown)


ios/AirshipReactNative.swift
```diff
@@ -328,10 +328,10 @@ public extension AirshipReactNative {
 // Actions
 @objc
 public extension AirshipReactNative {
-    func actionsRun(actionName: String, actionValue: Any?) async throws-> Any? {
+    func actionsRun(action: [String: Any]) async throws-> Any? {
         return try await AirshipProxy.shared.action.runAction(
-            actionName,
-            value: try AirshipJSON.wrap(actionValue)
+            action["_name"] as! String,
+            value: action["_value"] is NSNull ? nil : try AirshipJSON.wrap(action["_value"])
```
Ulrico972, line 334: I had to make sure I pass nil if it's an NSNull.
Otherwise the JSON wrapping fails: the NSNull Object don't go through the guard else, is that normal @crow ?
```
public static func wrap(_ value: Any?, encoder: JSONEncoder = AirshipJSON.defaultEncoder) throws -> AirshipJSON {
    guard let value = value else {
        return .null
    }
    .
    .
    .
}
```

jyaganeh (approved):
:+1:

crow (approved):

src/AirshipAnalytics.ts
```diff
@@ -33,15 +34,10 @@ export class AirshipAnalytics {
    * custom event is rejected.
    */
   public addCustomEvent(event: CustomEvent): Promise<null | Error> {
-    const actionArg = {
```
rlepinski, line 36: The proxy has a addCustomEvent method now that we can use instead of an action

src/AirshipActions.ts
```diff
@@ -9,15 +10,13 @@ export class AirshipActions {
   /**
    * Runs an Airship action.
    *
-   * @param name The name of the action.
-   * @param value The action's value.
+   * @param action The Airship Action.
    * @return A promise that returns the action result if the action
    * successfully runs, or the Error if the action was unable to be run.
    */
   public run(
```
rlepinski, line 17: You can avoid this breaking change by keeping the same interface that we have now and wrap the values to the module call:

```
return this.module.actionRun({"name": actionName, "value": actionValue});
```

src/Action.ts
```diff
@@ -0,0 +1,34 @@
...
+
+'use strict';
+
+import { JsonValue } from './types';
+
+/**
+ * Airship Action Object.
+ * This is used to encapsulate the Action name and the Action value.
+ */
+export class Action {
```
rlepinski, line 11: Instead of classes this should just be a simple interface with name/value, but we should be able to remove this

ios/AirshipReactNative.swift
```diff
@@ -328,10 +328,10 @@ public extension AirshipReactNative {
 // Actions
 @objc
 public extension AirshipReactNative {
-    func actionsRun(actionName: String, actionValue: Any?) async throws-> Any? {
+    func actionsRun(action: [String: Any?]) async throws-> Any? {
         return try await AirshipProxy.shared.action.runAction(
-            actionName,
-            value: try AirshipJSON.wrap(actionValue)
+            action["name"] as! String,
```
rlepinski, line 333: you will want to throw here instead of force unwrapping.

```
guard let name = action["name"] as? String else { throw AirshipErrors.error("missing name") }

File: src/CustomEvent.ts [hunk 6]
```diff
@@ -39,6 +41,34 @@ export class CustomEvent {
     this._transactionId = value;
   }
 
+  /**
+   * Gets the event's interaction ID.
+   */
+  get interactionId(): string | undefined {
```
rlepinski, line 47: do we need these changes?

rlepinski (approved):

PR #578, opened 2024-07-25
[MOBILE-4628] Add Webview inspection check in MessageCenter

Apekka:
### What do these changes do?
- Add a check in Message Center to allow the webview to be inspected in Safari
- Add a config field for isWebViewInspectionEnabled

To debug more easily the webviews

Putting isWebViewInspectionEnabled to true, running the sample app and inspecting the webview on Safari

image

To make testing faster and easier I edited the ProxyConfig so it loads the new field from the config, so I think it would be great to add it in the next Proxy update

image

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -481,11 +481,15 @@ export interface AirshipConfig {
      */
     itunesId?: string;
 
-
     /**
      * If set to `true`, the SDK will use the preferred locale. Otherwise it will use the app's locale.
      */
     useUserPreferredLocale?: boolean;
+
+    /**
+     * Allows the WebViews to be inspected in Safari.
+     */
+    isWebViewInspectionEnabled?: boolean;
   };
 
   /**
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -481,11 +481,15 @@ export interface AirshipConfig {
...
-
     /**
      * If set to `true`, the SDK will use the preferred locale. Otherwise it will use the app's locale.
      */
     useUserPreferredLocale?: boolean;
+
+    /**
+     * Allows the WebViews to be inspected in Safari.
+     */
+    isWebViewInspectionEnabled?: boolean;
```
rlepinski, line 492: this needs to be wired up in the mobile framework proxy

Apekka: Added in this PR : https://github.com/urbanairship/airship-mobile-framework-proxy/pull/46

Ulrico972 (approved):

PR #583, opened 2024-08-23
Make JsonObject accept undefined value

jp928:
### What do these changes do?

In the issue $ https://github.com/Expensify/App/issues/46820, the push notification data attribute extra may have value of undefined

Better type definition could result in less bug.

Send push notification to app without payload.


rlepinski:
👍

rlepinski:
Thanks!

PR #645, opened 2025-04-05
Release 23.1.0

rlepinski:
Lots of fun changes:
- Updates to latest template https://github.com/callstack/react-native-builder-bob
  - Drops commonjs due to the possibility of an app pulling our plugin in as both a module and a commonjs import. Since we have some state it would break things. 
  - Skipped the `import` change for the .h files on iOS. I believe those will only work in some circumstances and if it becomes an issue we will need to probably due an ifdef import for both paths
  - Added back `types` for apps that do not respect `exports` yet
  - Updated latest SDKs
  - Updated react native to 0.78.2


ios/MessageWebViewWrapper.swift
```diff
@@ -3,6 +3,7 @@
 import Foundation
 import AirshipKit
 import AirshipFrameworkProxy
+import WebKit
```
khmMouna, line 6: Do you need WebKit?

rlepinski: It references WKWebView within it so I think so. I dont know why it was working before

PR #738, opened 2026-05-09
Refactor docs publishing into reusable shared workflow

jyaganeh:
## Summary

- Extracts the `docs` and `deploy-docs` jobs from the release workflow into a new reusable workflow at `.github/workflows/publish-docs.yml`
- The shared workflow supports `workflow_call` (triggered automatically on tag push) and `workflow_dispatch` (manual republish for any existing tag), both checking out `ref: inputs.version`
- Updates `release.yml` to call the shared workflow via a single `publish-docs` job with `secrets: inherit`
- Upgrades GCP auth from deprecated `setup-gcloud@v0` to `google-github-actions/auth@v2` + `setup-gcloud@v2`
- Adds a `notify-release` job at the end of the release workflow for Slack notifications on success/failure/cancellation

## Test plan

- [ ] Verify `workflow_dispatch` trigger appears for `publish-docs.yml` in the GitHub Actions UI with a `version` input field
- [ ] Trigger a manual run against an existing tag and confirm docs build, GCS upload, and GitHub Pages deploy succeed
- [ ] Confirm the next tag-push release runs `publish-docs` and `notify-release` correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,74 @@
+name: Publish Docs
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+    secrets:
+      GCP_SA_KEY:
+        required: true
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag to publish docs for (e.g. 12.0.0)'
+        required: true
+        type: string
+
+jobs:
+  release-docs:
+    runs-on: ubuntu-latest
+    env:
+      NODE_ENV: development
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.version }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -0,0 +1,74 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.version }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+
+      - name: Setup GCP Auth
```
rlepinski, line 34: this is just temp right? dual deploy until its in?

rlepinski (approved):

.github/workflows/publish-docs.yml
jyaganeh: yeah. I'll circle back to remove all the old docs upload stuff once we've got everything publishing to pages

PR #741, opened 2026-05-20
[MOBILE-5612] Add sms/email registration

Apekka:
### What do these changes do?
Add SMS and Email registration to the React library, and to the example app

Test on the example app, Android and iOS, making sure SMS and Emails are registered

### DIFF
--- a/example/src/screens/HomeScreen.tsx
+++ b/example/src/screens/HomeScreen.tsx
@@ -244,17 +244,6 @@ export default function HomeScreen({ navigation }: HomeScreenProps) {
     }
   };
 
-  const registerSms = async () => {
-    setLoading(true);
-    try {
-      Airship.contact.registerSms("0669793554", )
-    } catch (error) {
-      // Ignore errors
-    } finally {
-      setLoading(false);
-    }
-  };
-
   useEffect(() => {
     const fetchInitialData = async () => {
       setLoading(true);
@@ -526,14 +515,14 @@ export default function HomeScreen({ navigation }: HomeScreenProps) {
               </View>
               <View style={styles.liveButtonRow}>
                 <Button
-                  title="Register SMS FR"
+                  title="Start New"
                   onPress={startLiveActivity}
                   color="#004bff"
                 />
               </View>
               <View style={styles.liveButtonRow}>
                 <Button
-                  title="Register Email"
+                  title="End All"
                   onPress={endAllLiveActivities}
                   color="#004bff"
                 />
--- a/example/src/screens/HomeScreen.tsx
+++ b/example/src/screens/HomeScreen.tsx
@@ -14,9 +14,11 @@ import {
   ActivityIndicator,
 } from 'react-native';
 import styles from '../Styles';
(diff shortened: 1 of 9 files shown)


khmMouna (approved):

example/src/screens/HomeScreen.tsx
```diff
@@ -203,6 +244,17 @@ export default function HomeScreen({ navigation }: HomeScreenProps) {
     }
   };
 
+  const registerSms = async () => {
+    setLoading(true);
+    try {
+      Airship.contact.registerSms("0669793554", )
```
crow: `registerSms` doesn't look like it's called from anywhere else, and the inner call has a trailing comma with no options arg. Wire it up or drop it?

example/src/screens/HomeScreen.tsx
```diff
@@ -414,14 +526,14 @@ export default function HomeScreen({ navigation }: HomeScreenProps) {
               </View>
               <View style={styles.liveButtonRow}>
                 <Button
-                  title="Start New"
+                  title="Register SMS FR"
```
crow: These two titles changed to "Register SMS FR" / "Register Email" (line 536 too), but onPress still points at `startLiveActivity` / `endAllLiveActivities`. Accidental swap?

src/types.ts
```diff
@@ -1114,3 +1114,27 @@ export interface LiveUpdateStartRequest {
...
+ */
+export interface EmailRegistrationOptions {
+  /** Milliseconds since epoch when the user opted into transactional emails. */
+  transactionalOptedIn?: number;
+  /** Milliseconds since epoch when the user opted into commercial emails. */
+  commercialOptedIn?: number;
+  /** Optional custom properties. */
+  properties?: Record<string, string>;
+  /** If true, a double opt-in confirmation email is sent. Defaults to false. */
+  doubleOptIn: boolean;
```
crow: Doc says "Defaults to false" but the field's required, so the default doesn't really apply. Make it optional?

```suggestion
  doubleOptIn?: boolean;
```

src/types.ts
```diff
@@ -1114,3 +1114,27 @@ export interface LiveUpdateStartRequest {
...
+ */
+export interface EmailRegistrationOptions {
+  /** Milliseconds since epoch when the user opted into transactional emails. */
+  transactionalOptedIn?: number;
+  /** Milliseconds since epoch when the user opted into commercial emails. */
+  commercialOptedIn?: number;
+  /** Optional custom properties. */
+  properties?: Record<string, string>;
+  /** If true, a double opt-in confirmation email is sent. Defaults to false. */
+  doubleOptIn?: boolean;
```
crow: oops, I think I got this one wrong last round. the iOS proxy at 15.9.0 has `doubleOptIn: Bool` non-optional. Putting it back the way it was.

```suggestion
  /** If true, a double opt-in confirmation email is sent. */
  doubleOptIn: boolean;
```

crow (approved):

PR #747, opened 2026-07-13
Add selection parameter to AirshipEmbeddedView

crow:
### What do these changes do?
Adds an optional `selection` parameter to `AirshipEmbeddedView` to control which pending instance displays when more than one is available for an embedded ID: priority (current default) or a specific instance ID. Also bumps the framework proxy to 15.11.0 to pick up the iOS and Android SDK versions that added this.

Both native SDKs added a way to target a specific pending embedded content instance instead of relying on priority ordering, and this exposes that to React Native apps.

Checked that the native parameter shapes generated on both platforms match what the bridge code reads, and ran typecheck against the diff (pre-existing unrelated errors on main are untouched). Added a demo card to the example app's Embedded Views screen exercising the new parameter.

### DIFF
--- a/android/src/main/java/com/urbanairship/reactnative/ReactEmbeddedViewManager.kt
+++ b/android/src/main/java/com/urbanairship/reactnative/ReactEmbeddedViewManager.kt
@@ -17,6 +17,8 @@ class ReactEmbeddedViewManager : SimpleViewManager<ReactEmbeddedView>(),
         override fun setProperty(view: ReactEmbeddedView, propName: String, value: Any?) {
             when (propName) {
                 "embeddedId" -> setEmbeddedId(view, value as? String)
+                "selectionType" -> setSelectionType(view, value as? String)
+                "selectionInstanceId" -> setSelectionInstanceId(view, value as? String)
                 else -> {}
             }
         }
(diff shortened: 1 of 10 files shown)


android/src/main/java/com/urbanairship/reactnative/ReactEmbeddedViewManager.kt
```diff
@@ -49,6 +51,16 @@ class ReactEmbeddedViewManager : SimpleViewManager<ReactEmbeddedView>(),
         }
     }
 
+    @ReactProp(name = "selectionType")
+    override fun setSelectionType(view: ReactEmbeddedView, selectionType: String?) {
```
rlepinski, line 55: Can we not pass in a json object that defines the selection?

src/RNAirshipEmbeddedViewNativeComponent.ts
```diff
@@ -1,8 +1,14 @@
 // @ts-ignore
 import { codegenNativeComponent, type HostComponent, type ViewProps } from 'react-native';
+import type {
+  WithDefault,
+  // @ts-ignore
+} from 'react-native/Libraries/Types/CodegenTypes';
 
 interface NativeProps extends ViewProps {
   embeddedId: string;
+  selectionType?: WithDefault<'priority' | 'instance_id', 'priority'>;
```
rlepinski, line 10: Can this be a json object?

PR #748, opened 2026-07-13
Release 26.8.0

crow:
Release 26.8.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # React Native Module 26.0.0 Changelog
 
+## Version 26.8.0 - July 13, 2026
+
+Minor release that updates the framework proxy to 15.11.0, picking up iOS SDK 20.10.0 and Android SDK 20.9.0.
+
+### Changes
+- Updated iOS SDK to [20.10.0](https://github.com/urbanairship/ios-library/releases/tag/20.10.0)
+- Updated Android SDK to [20.9.0](https://github.com/urbanairship/android-library/releases/tag/20.9.0)
+- Fixed native (Scenes) messages failing to render in the embedded message view
+- Added a `selection` parameter to `AirshipEmbeddedView` to pick which pending instance displays, by priority or a specific instance ID
+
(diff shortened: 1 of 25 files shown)


rlepinski (commented):
**`renderedConfig` set before the `embeddedId` guard** (`ReactEmbeddedView.kt`)

`renderedConfig = config` is assigned before the `embeddedId.isEmpty()` guard. If the first call delivers a valid JSON string with an empty `embeddedId`, `renderedConfig` captures it and every subsequent call with the same string short-circuits at the top — the view never initializes.

```kotlin
renderedConfig = config  // ← stored here

val embeddedId = json.opt("embeddedId").optString()
if (embeddedId.isEmpty()) {
    return  // ← but renderedConfig is already poisoned
}
```

Fix: move `renderedConfig = config` to after `addView(...)`.

---

**`ByInstanceId("")` vs `.priority` platform mismatch** (`ReactEmbeddedView.kt`)

When `type == "instance_id"` but `instanceId` is absent from the JSON, `optString()` returns `""` and Android calls `ByInstanceId("")` — which silently matches nothing and shows a blank view. iOS uses `if let instanceId = config.selection?.instanceId` and falls through to `.priority`.

```kotlin
AirshipEmbeddedSelection.ByInstanceId(selectionJson.opt("instanceId").optString())
//                                                                     ↑ "" when missing
```

Fix: mirror the iOS guard — only construct `ByInstanceId` when the string is non-empty, otherwise use `Priority`.

rlepinski (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # React Native Module 26.0.0 Changelog
 
+## Version 26.8.0 - July 13, 2026
+
+Minor release that updates the framework proxy to 15.11.0, picking up iOS SDK 20.10.0 and Android SDK 20.9.0.
```
rlepinski: We usually dont document the proxy version

crow: slop

PR #750, opened 2026-07-17
Update to RN 0.86

Apekka:
### What do these changes do?
Update to React Native 0.86

To say that our plugin is compatible with React Native 0.86, fixing important crashes

Running the sample app, making sure views are still working, and the headless JS service is functional

(ignore the 89 in the commit and branch name, it's actually 0.86)


rlepinski:
This looks good but the PR title and description should say 86 not 89

rlepinski (approved):

PR #755, opened 2026-08-14
Add Airship.getLaunchDeepLink for cold start deep links

crow:
### What do these changes do?
Adds `Airship.getLaunchDeepLink()`, a one-shot getter for the deep link that launched the app from a notification tap, wired for React Navigation's `linking.getInitialURL`.

Declarative-linking apps miss Airship deep links on cold start because the event fires after the navigation container mounts.

Typecheck and lint match main's baseline; builds and on-device cold start runs need the proxy release with `AirshipProxy.getLaunchDeepLink()` first.

### DIFF
--- a/src/AirshipRoot.ts
+++ b/src/AirshipRoot.ts
@@ -84,6 +84,11 @@ export class AirshipRoot {
    * Returns the deep link that launched the app from a notification tap, or
    * null if the app was not launched by a notification with a deep link.
    *
+   * Only reflects a cold app launch or background-to-foreground launch
+   * triggered by the tap. A notification tapped while the app is already in
+   * the foreground will not be returned here; use the `EventType.DeepLink`
+   * listener below for that case.
+   *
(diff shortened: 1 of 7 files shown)


src/AirshipRoot.ts
```diff
@@ -80,6 +80,34 @@ export class AirshipRoot {
     return this.module.isFlying();
   }
 
+  /**
+   * Returns the deep link that launched the app from a notification tap, or
+   * null if the app was not launched by a notification with a deep link.
+   *
+   * One-shot: the value is consumed on read. Intended to be wired as React
```
rlepinski, line 92: we should make it clear this only captures it from a cold launch

rlepinski (approved):

src/AirshipRoot.ts
crow: Good catch, updated the doc. Also think I found why CI's been red here, pushed a possible fix: https://github.com/urbanairship/airship-mobile-framework-proxy/pull/141

PR #760, opened 2026-09-04
[MOBILE-5857] Add feature flag status, waitRefresh, and status event

crow:
### What do these changes do?
Adds featureFlags.status(), waitRefresh(), and a FeatureFlagStatusChanged event across the JS and native layers.

Lets apps wait for fresh flag rules at launch and correct themselves when rules update mid-session (MOBILE-5857).

Typecheck and bob build pass, the Android library compiled against a locally published proxy/SDK chain, and the iOS example app built with the local proxy pod.

Red until the proxy releases with the new APIs, then the podspec and airshipProxyVersion get bumped here.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -174,6 +174,25 @@ export interface PushNotificationStatusChangedEvent {
 }
 
 /**
+ * Feature flag rule update status.
+ */
+export enum FeatureFlagUpdateStatus {
+  UpToDate = 'up_to_date',
+  Stale = 'stale',
+  OutOfDate = 'out_of_date',
+}
+
+/**
+ * Event fired when the feature flag rule update status changes.
+ */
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -173,6 +173,25 @@ export interface PushNotificationStatusChangedEvent {
   status: PushNotificationStatus;
 }
 
+/**
+ * Feature flag rule update status.
+ */
+export enum FeatureFlagUpdateStatus {
```
rlepinski, line 179: can you check this enum matches our other enums casing?

rlepinski (approved):

src/types.ts
crow: Checked, it matches the others, PascalCase keys with snake_case values like NotDetermined = 'not_determined'. The values also line up with what the proxy emits.

PR #762, opened 2026-09-15
[MOBILE-5920] Prepare release 26.13.0

Ulrico972:
### What do these changes do?
- skip explicit Kotlin plugin when AGP registers the kotlin extension - #761
- avoid build failure with react native 0.87

To make sure AGP9+ and react native 0.87+ usage don't break customers build

Manual test on the sample app. Reproduced the issue with AGP9 and react native 0.87, then verify it's fixing the issue.
Verified the fix doesn't break other AGP and react native versions too.


crow (approved):

Ulrico972:
CI is failing for now because we added feature flag status, waitRefresh, and status event in advance. The work is in review on the proxy side.

